PR #41, opened 2020-03-11
EM Command Line has been added. Fast load/spawn.

TicExp:
Ability to load/spawn with one click/command.
/cn - create a test room
/sp - spawn
click load/spawn to do both.

Its not final verison ofc.

### DIFF
--- a/Assets/Scripts/Extensions2.cs
+++ b/Assets/Scripts/Extensions2.cs
@@ -1,0 +1,91 @@
+﻿using UnityEngine;
+using System.Collections;
+using System.Text.RegularExpressions;
+
+public static class Extensions2
+{
+    private static readonly string Green = "[00ff00]";
+    private static readonly string Error = "[ff0000]";
+    private static readonly string Processing = "[ffa500]";
+    private static readonly string Warning = "[ffff00]";
+    private static readonly string Default = "[ffffff]";
+    private static readonly string Yellow = "[FFCC00]";
+    private static readonly string Red = "[FE2E2E]";
+
+    public static string SendCli(this string input)
+    {
+        EMCli.AddLine(input);
+        return input;
+    }
+
+    public static string SendError(this string input, bool bold = false)
+    {
+        EMCli.AddLine(input.RepaintError(bold));
+        return input;
+    }
+
+    public static string SendWarning(this string input, bool bold = false)
+    {
+        EMCli.AddLine(input.RepaintWarning(bold));
+        return input;
+    }
+
+    public static string SendProcessing(this string input, bool bold = false)
+    {
+        EMCli.AddLine(input.RepaintProcessing(bold));
+        return input;
+    }
+
+    public static string RepaintError(this string input, bool bold = false)
+    {
+        string colored = string.Concat(Error, input).hexColor2();
+        if (bold) return string.Concat("<b>", colored, "</b>");
+        return colored;
+    }
+    public static string RepaintWarning(this string input, bool bold = false)
+    {
+        string colored = string.Concat(Warning, input).hexColor2();
+        if (bold) return string.Concat("<b>", colored, "</b>");
+        return colored;
+    }
+    public static string RepaintGreen(this string input, bool bold = false)
+    {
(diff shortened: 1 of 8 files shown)


Assets/Scripts/Extensions2.cs
```diff
@@ -0,0 +1,91 @@
...
+        return input;
+    }
+
+    public static string SendProcessing(this string input, bool bold = false)
+    {
+        EMCli.AddLine(input.RepaintProcessing(bold));
+        return input;
+    }
+
+    public static string RepaintError(this string input, bool bold = false)
```
Lithrun: Repaint methods can be made private static

Assets/Scripts/Extensions2.cs
```diff
@@ -0,0 +1,91 @@
+﻿using UnityEngine;
+using System.Collections;
+using System.Text.RegularExpressions;
+
+public static class Extensions2
```
Lithrun: Don't name your class Extensions2. ConsoleExtensions would be a better name.

Assets/Scripts/UI/EMCli.cs
```diff
@@ -0,0 +1,184 @@
+﻿using UnityEngine;
+using System.Collections;
+using System;
+using Assets.Scripts.UI;
+using UnityEngine.SceneManagement;
+
+public class EMCli : MonoBehaviour
+{
+    private static Rect windowRect = new Rect(Screen.width - 320f, 0, 300f, 430f);
+    private static GUILayoutOption[] options1 = new GUILayoutOption[] { GUILayout.Width(320f), GUILayout.MaxHeight(430f) };
```
Lithrun: Rename property to options. There's just options, so why should we name it options1?

Assets/Scripts/UI/EMCli.cs
```diff
@@ -0,0 +1,184 @@
...
+    private static GUI.WindowFunction func1 = new GUI.WindowFunction(WindowLayout);
```
Lithrun: Same as above, but name it windowFunction as func could be confused with Func<T1, TOut>

Assets/Scripts/UI/EMCli.cs
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+
+    void OnGUI()
+    {
+        if (Visible)
+        {
+            if (Event.current.type == EventType.KeyUp && Event.current.keyCode == KeyCode.RightAlt)
+            {
+                if (GUI.GetNameOfFocusedControl() == "CLI")
```
Lithrun: Avoid using hardcoded strings multiple times. Move "CLI" within one private static readonly "CLI"

Assets/Scripts/UI/EMCli.cs
```diff
@@ -0,0 +1,184 @@
...
+                $"[{command}] - there is no such command!".SendError(true);
+            }
+        }
+        else
+        {
+            $"[{command}] is not a command!".SendError(true);
+        }
+    }
+
+    public static IEnumerator ConnectAndJoinIE(bool spawnAfterLoad)
```
Lithrun: This method is nice, but it does throw ~200 error messages "JoinOrCreateRoom failed. Client is not on Master Server or not yet ready to call operations. Wait for callback: OnJoinedLobby or OnConnectedToMaster.
UnityEngine.Debug:LogError(Object)". I assume that this is just temporary?

Assets/Scripts/UI/EMCli.cs
```diff
@@ -0,0 +1,184 @@
...
+    {
+        GUI.backgroundColor = Color.green;
+        GUILayout.BeginHorizontal();
+        if (GUILayout.Button(">", GUILayout.MaxWidth(25)))
+        {
+            try
+            {
+                EnterCommand(inputLine);
+            }
+            catch (Exception ex)
```
Lithrun: Rather catch a specific exception here. Commands will execute game logic. If the game logic has an exception, it should also appear within the unity error log.

Lithrun:
PR closed as working via issue #43 now.

PR #68, opened 2020-04-05
#58 cli enhancements

TicExp:
CLI enchancements. First edit.

PR #78, opened 2020-04-11
#68 cli enhancements

TicExp:


Assets/Scripts/UI/CommandLine/EMCli.cs
```diff
@@ -3,62 +3,171 @@
...
 
-    public static void ClearLayout()
+    public static void RefreshLayout()
     {
         layout = string.Empty;
+        foreach(ConsoleMessage message in Messages)
+        {
+            foreach(DebugLevelOption option in DebugLevelOption.DebugLevelOptions)
+            {
+                if (option.DebugLevel.Equals(message.ConsoleMessageDebugLevel))
```
Lithrun, line 136: Why a double if statement here? Just do
```suggestion
                if (option.DebugLevel.Equals(message.ConsoleMessageDebugLevel) && option.Enabled)
```

Issue #145, opened 2020-05-19
[BUG] Visible Akina Racing checkpoints do not restore gas


SlowDeepExhalation:
**Describe the bug**
The floating green gas canisters in Akina racing do not restore gas.

**To Reproduce**
1. Start an Akina Racing server.
2. Proceed through the race until you find a green canister.  Enter the canister and note that your gas is not restored.

**Expected behavior**
Your gas should be restored upon entering any of the green canisters, and then the canister should disappear and become unusable.

**Additional context**
Build: Stresstest

SlowDeepExhalation:
Fixed in build144-145

Issue #144, opened 2020-05-19
[BUG] Invisible Akina Racing checkpoints restore gas 


SlowDeepExhalation:
**Describe the bug**
The invisible checkpoints in Akina racing restore gas, in addition to setting the player's spawnpoint.  They also send a green "Checkpoint set." message to the chatbox, just like custom racing checkpointCuboids.

**To Reproduce**
1. Start an Akina Racing server.
2. Go through the course until you reach the lava sea.  This is just past the first invisible checkpoint.  You should have full gas, and a green message in your chatbox.  You can gasburst a few times then go backwards to repeat it.

**Expected behavior**
As in RCmod, the invisible checkpoints should not restore gas or give a message that a checkpoint has been set.  They should only set your respawn point.

**Additional context**
The checkpoints do set your respawn point, at least.
Build: Stresstest

SlowDeepExhalation:
Fixed in build144-145

Issue #83, opened 2020-04-12
New Gamemode


KoroSensei0320:
New game mode suggestion: checkpoint racing 

Checkpoint racing is a new game mode, which opens up a lot of new opportunities and options for racers and racing map makers. Whereas in the regular racing game mode, you'd have to go from start to finish, with checkpoint racing players will have to go from checkpoint to checkpoint to checkpoint. You will go from checkpoint A, to B, to C, to D (ect. ect. ect.), where the final checkpoint acts as the finish line. In order to finish, you need to sequentially go through the checkpoints (A-B-C-D ect.).

**Additional context**
May need to discuss this game mode further in the Development Squad channel, or investigate for more information.

Lithrun:
Test

Lithrun:
Renaming this issue to Racing Objectives, to prevent confusion in terminology with the green checkpoints. Players are required to pass through Racing Objectives in chronological order in order to complete a race. Racing Checkpoints are for gas resupply and respawn checkpoints

Lithrun:
Current implementation:
- A new racing prefab is introduced, named "Racing Objective". A racing objective is a ring, in which players have to ODM through.
- Going through a racing objective will set a checkpoint and refill gas.
- Going through a racing objective, will popup an arrow for 5 seconds into the direction of the next objective
- When a racing objective does not have a next objective, this is seen as the final checkpoint
- Racing objective has three states: Current, Inactive and Passed. The player has to go through a "Current" objective, which is colored via Blue and Gold. Objectives that were already completed, are given the color of grey. Objectives that are yet to be crossed (Inactive), are marked with Grey and Red, to indicate that they are still locked.

PR #147, opened 2020-05-19
#144 #145 akina checkpoint and gas bugs

DjFrytek:
Fixed checkpoints restoring gas and printing a message.
Created new script for filling gas on trigger and attached it to gas restore points on Akina Racing map.


Assets/Scripts/Gamemode/Racing/RacingCheckpointTrigger.cs
```diff
@@ -11,8 +11,6 @@ private void OnTriggerEnter(Collider other)
             gameObject = gameObject.transform.root.gameObject;
             if ((((IN_GAME_MAIN_CAMERA.gametype == GAMETYPE.MULTIPLAYER) && (gameObject.GetPhotonView() != null)) && gameObject.GetPhotonView().isMine) && (gameObject.GetComponent<Hero>() != null))
             {
-                FengGameManagerMKII.instance.chatRoom.addLINE("<color=#00ff00>Checkpoint set.</color>");
```
Lithrun: If we hit a checkpoint, we still want to display that we have a checkpoint set.

Lithrun (changes requested):
Rework the single comment, then create the builds for testing team.

Lithrun:
Upon further investigation I've concluded that the changes to RacingCheckpointTrigger have to be reverted and that RacingGasTrigger can be deleted. Whereas this does fix the bugs, it will break custom racing maps.

The RacingCheckPoint trigger is intended for the Map Editor checkpoints, as going through a checkpoint of a custom racing map, gives you gas and sets your checkpoint. Because of this, the RacingCheckpoint has to be reverted back or we will break custom racing maps that require this.

For Akina itself, we should use the "LevelTriggerCheckpoint" class, as all this class does is assigning a checkpoint, which is the required behavior. "LevelTriggerGas" should be used for objects which give gas, and are then expected to be deleted.

Introducing additional triggers to the racing might be useful, but that's not required to fix this bug. There's room for refactoring possible, but save that for issue #83 as in order to add new racing features, we need some cleanup

Lithrun (approved):

SlowDeepExhalation (approved):

Issue #130, opened 2020-05-12
[BUG] In-game cursor disappears 


KoroSensei0320:
**Describe the bug**
The ingame cursor disappears in **Test Zone ** if you change camera. Same problem arises if you enter settings and then exit.

**To Reproduce**
Steps to reproduce the behaviour:
1. Host a server using any map.
2. Press C to change camera OR enter then exit settings.
3. See error - this hides the crosshair and shows your OS cursor. The cursor still works to fire hooks.

**Expected behavior**
It is expected that the crosshair shows up visually for all camera modes.

**Log file**
[output_log.txt](https://github.com/AoTTG-2/AoTTG-2/files/4612761/output_log.txt)

Additional info: 
The crosshair constantly works for TPS, but only for that camera view.

JupiterJaeden:
Upon further study, it seems we are dealing with 2 separate (although not entirely unrelated) bugs. 

Bug 1: Opening the menu makes hardware cursor visible, as expected. However, closing the menu causes the in-game reticle to disappear and hardware cursor remains visible. 
Bug 2: Switching camera modes causes in-game reticle to disappear and hardware cursor to become visible. However, switching to the TPS camera mode fixes this. 

I will attempt to solve both of these bugs, starting with bug 1.

PR #158, opened 2020-05-24
#130 cursor disappearing bug

JupiterJaeden:
Alright, so basically I identified that this was actually 2 bugs, not 1. 

To solve bug 1 I edited the InGameMenu.cs and InGameUI.cs files. I moved some stuff out of the InGameUI.cs file and consolidated it in the InGameMenu.cs file. You can see my changes there.

To solve bug 2, I edited the IN_GAME_MAIN_CAMERA.cs file. I removed some old obsolete stuff that was causing the problem and replaced it with up to date stuff. You can see my changes there.


JupiterJaeden:
Oh and btw this has already been playtested, everything does work as intended.

JupiterJaeden:
Do not approve this pull request, I have just discovered that my solution has caused another (albeit minor) bug. Basically, upon pressing the "Quit" button in the in-game menu and returning to the main menu, the cursor is no longer visible. This should be easily fixable and I will go ahead and do that right now.

PR #173, opened 2020-05-28
Reassign attack_kick animations

mwnDK1402:


mwnDK1402:
I tested it by letting myself get kicked to death. Can confirm; the titans do kick now.

PR #189, opened 2020-06-12
Implement SuperUnityBuild

### LINKED ISSUE #188: Better Build Tools
I'm on the @AoTTG-2/development-team, and I found that it took me about an hour, probably more, to make a proper build for the first time. Part of this is because of switching platforms, and having to keep checking whether it's done doing what it's doing. I would like to find some way to make building the game faster and more convenient.

I went looking for tools, and this is an open-source project I found, which I don't believe has any drawbacks for us: https://github.com/Chaser324/unity-build
As far as I can tell, it will allow us to build for the platforms we want with the click of a button. Maybe we can find some way to upload automatically too!

mwnDK1402:
Closes #188


mwnDK1402:
Problems so far:
- I haven't found a way to set the compression to Lz4HC.
- There seems to be a problem with saving the assets properly, which can be seen in Version Control. Seems like pressing `Ctrl + S` in the scene has an effect on whether changes to the assets are saved, even though it shows no unsaved changes.

mwnDK1402:
Seems like it works pretty well, but I kept the UI simple by cutting out a lot of customization and replacing it with references to `VersionManager`. I just thought of a different way of doing it, which is just storing the version in `PlayerSettings.bundleVersion` instead of in `VersionManager`. The purpose of the `VersionManager` is to compute the version from the branch name. This should be done before the game runs, so that the correct version is displayed when playing in the editor. If we use `PlayerSettings.bundleVersion` instead of a `ScriptableObject`, we could do this in a different editor window. Maybe just the UnityBuild editor.

Lithrun:
Closing this PR as it has been open since June without updates. Re-open once ready for Review.

PR #191, opened 2020-06-18
#166 chat enhancements

### LINKED ISSUE #166: [UI] Chat Enhancements
The current chat within AoTTG2 is functional, but there's some improvements we can apply.
- Chat Scrollbar. Allow players to scroll up to look up previous messages. Cache up to 100 messages.
- Chat Character Limit. Ever wanted to copy pasta an entire book into AoTTG? Well, you can. Implement a character limit on chat messages so that not an entire chat can be spammed with a single wall of text. Make this limit 1000.
- Chat formatting. When players are using a broken format tag, like "<color" without properly closing it, it will bug out the entire chat. Apply a regex to fix this
-Inputs. When players type in chat the player inputs are still active as they are typing. Disable player inputs while typing.

A chat system with some quality of life additions.

- [x] Implement a scrollbar for chat, and limit the amount of messages that are stored inside the InRoomChat.cs to 100.
- [x] Add a character limit to chat messages.
- [x] Using broken tags should not break the chat.
- [x] Player inputs disabled while chat is focused.

Zoap:
I took the liberty of adding to the fixes in this issue even tho I'm not supposed to (I don't like leaving things broken) and added "Disable player input  when chat is active" as one of the issues to fix in this Issue.

Summary of what has been done:
- Total refactoring of InChatRoom.cs (Resulted in three new classes)
- ChatCommands.cs enum for the chat commands available
- ChatCommandHandler.cs handles everything to do with chat commands and their logic
- ChatUtility.cs static class with methods for formatting chat text

- [x] Implement a scrollbar for chat, and limit the amount of messages that are stored inside the InRoomChat.cs to 100.
- [x] Add a character limit to chat messages.
- [x] Using broken tags should not break the chat.
- [x] Player inputs disabled while chat is focused.

Closes [#166](https://github.com/AoTTG-2/AoTTG-2/issues/166)


Zoap:
Something I forgot to mention was that the chat is scrollable, but I couldn't for the life of me figure out how to get the scrollbar working, so I never added the scrollbar and was going to pass that to someone else, but as you see I forgot. So... how important is the actual scrollbar?

Assets/Scripts/Room/Chat/ChatCommandHandler.cs
```diff
@@ -0,0 +1,777 @@
...
+        foreach (var ignoredPlayer in ignoreList)
+        {
+            var message = $"{ignoredPlayer}";
+            instance.chatRoom.AddMessage(message);
+        }
+    }
+
+    private static void OutputIsRc()
+    {
+        string message;
```
mwnDK1402:
```cs
string message = masterRC ? "Is RC" : "Not RC";
```

Assets/Scripts/Room/Chat/ChatCommandHandler.cs
```diff
@@ -0,0 +1,777 @@
...
+            {
+                switch (teamInt)
+                {
+                    case 0:
+                        SwitchTeam(0);
+                        message = "You have joined individuals.";
+                        break;
+                    case 1:
+                        SwitchTeam(1);
+                        message = FormatTextColor00FFFF("You have joined team cyan.");
```
mwnDK1402: Should be able to find a more intuitive way to express this. With an enum like `enum Team { None, Cyan, Magenta }`, for example. Or with a dictionary.

Assets/Scripts/Room/Chat/ChatCommandHandler.cs
```diff
@@ -0,0 +1,777 @@
...
+    /// </summary>
+    /// <param name="chatCommand"></param>
+    public static void CommandHandler(string chatCommand)
+    {
+        var commands = chatCommand.Replace("/", "").ToLower().Split(' ');
+        var parameter = string.Empty;
+
+        ChatCommand command;
+
+        if (Enum.TryParse(commands[0].ToLower(), out command))
```
mwnDK1402: I think I would prefer a `Dictionary<string, CommandHandler> commandHandlerByName`. Would make this code as simple as this:
```cs
CommandHandler handler;
if (TryGetCommandHandler(command, parameter, out handler) && HasPermissionToUse(handler))
    handler.Execute();
```
... where there would be a `Dictionary<name, CommanHandler> commandHandlerByName` used by `TryGetCommandHandler`.

Assets/Scripts/Room/Chat/InRoomChat.cs
```diff
@@ -1,593 +1,229 @@
...
+        else
+        {
+            OutputErrorMessage($"Message can not have more than {MaxMessageLength} characters");
+        }
+    }
+
+    /// <summary>
+    /// Removes messages when exceding max storage
+    /// </summary>
+    private void RemoveMessageIfMoreThenMax()
```
mwnDK1402: Typo here: `MoreThanMax`.
How about something like `TrimMessages`?

Assets/Scripts/Room/Chat/ChatCommandHandler.cs
```diff
@@ -0,0 +1,777 @@
...
+            {
+                var playerToBan = playerList.FirstOrDefault(p => p.ID == playerId);
+
+                if (!playerList.Any(p => p.ID == playerId))
+                {
+                    instance.chatRoom.OutputErrorPlayerNotFound(playerIdString);
+                    return;
+                }
+
+                if (OnPrivateServer || isMasterClient)
```
mwnDK1402: This check seems redundant, since above is an `if (!A)`, and this is checking `if(A)`.

Assets/Scripts/Room/Chat/ChatCommandHandler.cs
```diff
@@ -0,0 +1,777 @@
...
+            instance.chatRoom.OutputErrorMessage("Fov has to be a number");
+        }
+    }
+
+    /// <summary>
+    /// Enter/Exit spectate mode
+    /// </summary>
+    private static void EnterExitSpecMode()
+    {
+        string message;
```
mwnDK1402: Rename to `ToggleSpecMode()`.
```cs
// This will be more readable when we improve the settings.
settings[0xf5] = ((int)settings[0xf5] == 1 ? 0 : 1);
bool specMode = settings[0xf5] == 1;
instance.EnterSpecMode(specMode);
string message = FormatSystemMessage(specMode ? "You have entered spectator mode." : "You have exited spectator mode.");
instance.chatRoom.AddMessage(message);
```

Assets/Scripts/Room/Chat/ChatCommandHandler.cs
```diff
@@ -0,0 +1,777 @@
...
+        {
+            instance.chatRoom.OutputErrorNotMasterClient();
+        }
+    }
+
+    /// <summary>
+    /// Send private message to player with Id
+    /// </summary>
+    /// <param name="input"></param>
+    private static void SendPrivateMessage(string[] input)
```
mwnDK1402: This seems super confusing. `string[] input` is just the list of parameters. And we string them all back together, after they've been split up. :P
So `input` should be renamed to `parameters`. `whatIsThisVar` is the actual text pf the message to be sent. Maybe `messageText` or `rawMessage`.
Maybe we should use a `StringBuilder` here.

Assets/Scripts/Room/Chat/ChatCommandHandler.cs
```diff
@@ -0,0 +1,777 @@
...
+            ChatCommand command;
+            if (input.Count() >= 3)
+            {
+                if (Enum.TryParse(input[1], out command))
+                {
+                    string parameter = input[2];
+                    int maxPlayers;
+                    float time;
+
+                    switch (command)
```
mwnDK1402: Same as L662. This one hasn't even been turned into methods, though.

Assets/Scripts/Room/Chat/InRoomChat.cs
```diff
@@ -1,593 +1,229 @@
...
     private string inputLine = string.Empty;
     public bool IsVisible = true;
-    public static List<string> messages = new List<string>();
-    private Vector2 scrollPos = Vector2.zero;
+    private readonly List<string> messages = new List<string>();
     public InputField ChatInputField;
     public Text ChatText;
 
-    public void addLINE(string newLine)
+    public void OnGUI()
```
mwnDK1402: Is this the correct Unity Message to use?
https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnGUI.html

mwnDK1402 (commented):
I tried just looking through it and seeing what things should probably be refactored.

Assets/Scripts/Room/Chat/ChatCommandHandler.cs
Zoap: It's a great solution, however, I do believe that we can do that refraction at a later date, maybe when we start adding more commands we could add this as an issue to be resolved, but for administrative purposes I don't think we should resolve it in this branch.

Assets/Scripts/Room/Chat/InRoomChat.cs
Zoap: I have no idea.

Assets/Photon Unity Networking/UtilityScripts/PunTeams.cs
```diff
@@ -102,7 +102,7 @@ public static PunTeams.Team GetTeam(this PhotonPlayer player)
         object teamId;
         if (player.CustomProperties.TryGetValue(PunTeams.TeamPlayerProp, out teamId))
         {
-            return (PunTeams.Team)teamId;
+            return (PunTeams.Team)(int)teamId;
```
Lithrun, line 105: Is there a reason why this change is introduced? Photon Unity Networking is an unity asset (third party library). We shouldn't modify this.

Assets/Scripts/Room/Chat/ChatCommandHandler.cs
```diff
@@ -0,0 +1,722 @@
...
+                }
+            }
+        }
+    }
+
+    /// <summary>
+    /// Revive player with Id
+    /// </summary>
+    /// <param name="playerIdString"></param>
+    private static void RevievePlayer(string playerIdString)
```
Lithrun: *Revive

Assets/Scripts/Room/Chat/ChatCommandHandler.cs
```diff
@@ -0,0 +1,722 @@
...
+        hashTable.Add(PhotonPlayerProperty.total_dmg, 0);
+        player.SetCustomProperties(hashTable);
+
+        var message = "Your stats have been reset.";
+        instance.chatRoom.AddMessage(message);
+    }
+
+    private static void ChangeRoomProperties(string[] input)
+    {
+        if (isMasterClient)
```
Lithrun: There's several if statements here. Code can be made more readable with
```
if (!isMasterClient)
{
    instance.chatRoom.OutputErrorNotMasterClient();
    return;
}

// Now we can assume that we are the master client
```

The remaining code has to be in one less if statement.

Assets/Scripts/Room/Chat/ChatCommands.cs
```diff
@@ -0,0 +1,35 @@
+﻿public enum ChatCommand
```
Lithrun, line 1: Enums should start with a capital letter, following PascalCase

Lithrun (changes requested):
ChatCommandHandler is looking neat so far! A few code improvement tips in review 👍

Assets/Photon Unity Networking/UtilityScripts/PunTeams.cs
Zoap: L56 in ChatUtility.cs, the player.GetTeam() method I assumed should return a PunTeams.Team enum, but it didnt until I made the cast in that method, but I see now that I maybe should make the cast outside of method instead. I'll change it.

Zoap: InvalidCastException: Specified cast is not valid.
TeamExtensions.GetTeam (PhotonPlayer player) (at Assets/Photon Unity Networking/UtilityScripts/PunTeams.cs:105)
ChatUtility.SetNameColorDependingOnteam (PhotonPlayer player) (at Assets/Scripts/Room/Chat/ChatUtility.cs:56)
InRoomChat.ChatAll (System.String message) (at Assets/Scripts/Room/Chat/InRoomChat.cs:77)
InRoomChat.HandleChatInput (InRoomChat chat) (at Assets/Scripts/Room/Chat/InRoomChat.cs:116)
InRoomChat.OnGUI () (at Assets/Scripts/Room/Chat/InRoomChat.cs:30)

Zoap: It seems that casting object to PunTeams.Team is not valid for some reason

Lithrun: Let's see. Perhaps the problem is that the PunTeams.Team is of type byte and not of type integer?

Zoap: Yeah but in my switch case I am using Punteams.Team

```
var playerTeam = player.GetTeam();
switch (playerTeam)
{
case PunTeams.Team.red:
    name = FormatTextColor00FFFF(name);
    break;
case PunTeams.Team.blue:
    name = FormatTextColorFF00FF(name);
    break;
default:
    break;
}
```

Assets/Scripts/Room/Chat/InRoomChat.cs
mwnDK1402:
```suggestion
    private void Update()
```
> OnGUI is the only function that can implement the "Immediate Mode" GUI (IMGUI) system for rendering and handling GUI events. Your OnGUI implementation might be called several times per frame (one call per event).

Lithrun (approved):

PR #200, opened 2020-06-20
#182 default textured houses

### LINKED ISSUE #182: [BUG] Default textured houses
**Describe the bug**
When opening the City and Outside the Walls scene, for some reason the houses do no longer have textures applied to them.

**To Reproduce**
1. Open the City / OutsideTheWalls scene in the unity editor
2. Houses do not have textures and are white

**Expected behavior**
It is expected that houses have a texture and do not look white

**Screenshots**

**Log file**
Include a log file

**Additional context**
image

Zoap:
- Fixed texture issues on houses, walls and ground on scenes: Colossal, The City I and Outside
- House material was missing texture
- Walls used wrong material

Known Issue:
Gas resupply station on on City I contaminates wall and ground with wrong material, whatever material set to aot_supply gets loaded to wall and ground when the scene is loaded. I set the wall material/texture to the aot_supply as a temporary solution so the supply station will have the wrong texture.

Closes #182


Zoap:
I'll resolve the conflicts when I'm done fixing #191, unless you can resolve them @Lithrun

Lithrun (approved):

PR #212, opened 2020-06-30
#207 Fixed the crawler eyes ( cant be blind now)

### LINKED ISSUE #207: Disabling Crawler Eye cutting
Currently, players can damage the eyes of crawlers. This should not happen, since defeating crawlers is very simple because of this small oversight.

Humans should not be able to cut the eyes of crawlers.

- [x] Navigate to a class named "MindlessTitan", and modify the OnEyeHitRpc, so that nothing will happen if the current MindlessTitan is of type: MindlessTitanType.Crawler.

Revaneuh:


Lithrun (approved):
Crawler eyes can no longer be damaged. Nice one 👍

PR #233, opened 2020-07-19
#179 reel in

### LINKED ISSUE #179: [BUG] Reel-in feels unresponsive, especially with default controls
**Describe the bug**
Reeling in feels unresponsive to many testers, including some who are used to playing with default controls.  The responsiveness seems to go up when performing critical reel ins (reeling in the moment a hook lands), and goes down when orbiting around an object; this may be related to issue #97 .

Rebinded reel in using AutoHotKey (which causes another key to send many WheelDown events, as if the player were scrolling the mousewheel extremely fast) *seems* to be more reliable than normal mousewheel scrolling, but some still report unresponsiveness.  Since rebinds *aren't actually implemented*, it's impossible to determine whether the rebinded unresponsiveness is due to AHK or AoTTG 2.

**To Reproduce**
The bug has no true reproduction method; testers *feel* like it's wrong, but it's not nonfunctional, just unreliable.  So, play forest and reel between trees.  Pay attention to the timing of successful reels.  It often feels late.

**Additional comments**
Confirmed on Issue98

May be related to the old AoTTG bug where higher FPS causes reeling to be less responsive because it has to happen on one of the frames where the game actually checks input.  We cannot easily compare how the game performs at different FPS rates due to Issue #178
May be related to Issue #97, since critical reels work better.
*Please* implement rebinds.  The current situation is just terrible.

miguelgmarques:
Wrote a update Method where it checks the input for reel binds


Assets/Scripts/Characters/Human/Hero.cs
```diff
@@ -181,6 +181,8 @@ public class Hero : Human
     private float uTapTime = -1f;
     private bool wallJump;
     private float wallRunTime;
+    private bool ReelInInput=false;
```
Lithrun: The default value of a bool is false. Thereby it's not required to state that they are false.

Assets/Scripts/Characters/Human/Hero.cs
```diff
@@ -1098,6 +1100,14 @@ private void AddRightForce()
         baseRigidBody.AddForce(baseRigidBody.velocity * 0.00f, ForceMode.Acceleration);
     }
 
+    private void Update(){
```
Lithrun, line 1090: Code looks alright, but please apply C# syntax style. This looks more Java than C#. Pressing Ctrl + K, D, should do this automatically

Lithrun (changes requested):

Lithrun:
Code in terms of functionality is good. However, maintainability could use some tweaks. See comments above. Additionally, please undo the changes to the `AoTTG2-sln`.

Additionally, please create 3 builds for testing team as described here: https://github.com/AoTTG-2/AoTTG-2/wiki/How-to-create-a-new-testing-build

Lithrun:
As mentioned from Testing Team, the change doesn't appear to solve the problem. So please re-investigate an alternative approach

Lithrun:
Closing this PR as it has been open since August without updates. Re-open once ready for Review.

PR #252, opened 2020-08-28
#206 restart/suicide key

### LINKED ISSUE #206: [BUG] Restart / Suicide key
**Describe the bug**
Upon using the restart key as a human, the player will not be able to kill themselves. This is caused by an exception related to weapon trails. Additionally, player titans do not have the functionality of the Restart / Suicide Key implemented. Although that may technically not be considered as a bug, please include that within this issue as both are related to the same key.

**To Reproduce**
1. Spawn as a Human or Titan (Enable PlayerTitans in order to reveal this option at the spawn menu)
2. Pressing the "Restart" key (Default is T. Check: Controls > UI > Restart for your rebind)
3. Nothing happens

**Expected behavior**
It is expected that a player can kill themselves whether they be a human or titan.

- [x] On Hold until #61 is completed

FizzySD:
Closes #206


Assets/Scripts/Characters/Human/Hero.cs
```diff
@@ -4543,10 +4559,14 @@ public IEnumerator stopImmunity()
     }
 
     private void suicide2()
-    {
-        this.netDieLocal((Vector3) (base.GetComponent<Rigidbody>().velocity * 50f), false, -1, string.Empty, true);
-        FengGameManagerMKII.instance.needChooseSide = true;
-        FengGameManagerMKII.instance.justSuicide = true;
+    {      
+        if (!PhotonNetwork.offlineMode)
```
Lithrun: Using `if (PhotonNetwork.offlineMode) return` would be preferred here rather than nesting your code within an if statement. Additionally, suicide2() is currently only called from this code:
`                                if (!PhotonNetwork.offlineMode)
                                { 
                                this.suicide2();
                                }
`

Please remove the if statement check from this method.

Assets/Scripts/Characters/Titan/PlayerTitan.cs
```diff
@@ -164,6 +164,27 @@ protected override void OnTitanDeath()
             PhotonNetwork.player.SetCustomProperties(propertiesToSet);
         }
 
+        public void ptdie()
```
Lithrun: ptdie isn't a proper method name. Additionally we are using PascaleCase in C#, as this is not Java or Python. Rename the method to "Die", as we are already within the Player Titan class, and currently you are saying "Player Titan. Player Titan Die" whereas "Player Titan.Die" makes more sense.

Assets/Scripts/UI/Camera/SpectatorMovement.cs
```diff
@@ -25,18 +33,24 @@ private void Update()
             {
                 num2 = -1f;
             }
+
             else
             {
                 num2 = 0f;
             }
+            if (Input.GetKeyDown(KeyCode.T))
```
Lithrun: Code below shows that InputManager is used, so this if statement can be removed

Assets/Scripts/UI/Camera/SpectatorMovement.cs
```diff
@@ -6,13 +6,21 @@ public class SpectatorMovement : MonoBehaviour
     public bool disable;
     private float speed = 100f;
 
+    private void reset()
```
Lithrun: Name should be `reset`. Please see: https://www.dofactory.com/reference/csharp-coding-standards

Assets/Scripts/UI/Camera/SpectatorMovement.cs
```diff
@@ -50,6 +64,10 @@ private void Update()
             {
                 transform.position -= (Vector3) ((base.transform.forward * speed) * Time.deltaTime);
             }
+            if (InputManager.KeyDown(InputUi.Restart))
```
Lithrun, line 64: Seems like here you are using InputUi.Restart. Do note, that the "this" modifier is not required and should rather be avoided in small classes.

ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 2017.4.40f1
```
Lithrun: Our Unity version is 2017.4.40f1, please do not downgrade the project.

Lithrun (changes requested):
Overall its good, but a few maintainability improvements required.

Lithrun (approved):

Issue #277, opened 2020-09-11
[BUG] Racing Timers


DjFrytek:
**Describe the bug**
The timer shown in the middle of the screen after finishing the race keeps going even though the race is over, furthermore it adds additional 15 seconds. Also the timer on top sometimes shows numbers like 21.000003.

**To Reproduce**
1. Go to 'City racing mode'
2. Finish the race
3. See the timer in the middle keep going

**Expected behavior**
The timer in the middle should stop when the finish line is reached and display the time it took the player to finish. The time should be counted from the moment the player is released from the spawn as the timer on the top of the screen.
The top timer also should be restricted to 1 digit after the decimal place, so it doesn't show numbers like 21.000003.

**Screenshots**
![](https://cdn.discordapp.com/attachments/683074276450238519/754029558763749497/unknown.png)

**Additional context**
Despite the timer in the middle still going, after restarting the map the saved record is correct. The possible reason the timer is 15 seconds ahead, is the fact that in orginal akina racing the preparation time is 5 seconds, while in the city racing it's 20 seconds, meaning the middle timer might be counting starting at 5 seconds after the map is loaded, instead of 20.

**Contributor instructions** (Do not modify)
1. Try to recreate the bug via the reported reproduction steps. If the bug can be reproduced, remove the investigation label. If the bug cannot be reproduced, please add a comment, attach "On Hold" label and inform the submitter that the bug cannot be reproduced. If there is no reply after 2 weeks, the issue can be closed.
2. Follow the default development team procedures (Create a new branch, apply your fix, create a pull request, submit new builds for testing team in dev squad)
3. Once testing team has confirmed the issue is resolved and the pull request is accepted, the issue can be closed.

Mi-Sad:
unfortunately the code i have to rework to fix the issue is quite a chunck of code as it's into the spaghetti FGMMKII core2 which as is setted as obsolete i guess it has to be totally reworked if no active issue include redoing it imma rework the game mode system in order to create a default virtual method for gamemode and customize the method in each gamemode that require it (basically UI in game info management)

Lithrun:
#160 is completed, so feel free to finish this issue

Mi-Sad:
Roger ;)

Lithrun:
For this issue you do require to modify code of FengGameManager. Where possible, please move any methods / properties marked with [Obsolete] which contain a message that this method / property should be migrated to RacingGamemode.

[...]

PR #282, opened 2020-09-18
#277 Racing Timer

Mi-Sad:
Added string formatting to room timer so that no weird .0003 would pop up in the timer (racing) and also added some component to actually calcultate the last local result and store it in the var inside racinggamemode and also display it, i honeslty don't like this solution but works perfectly for the moment (as what i like would require a total rework of the FGMMKII class which is in a hold state till some other issue are solved)

Here's a video of how it behave (video not public)
https://youtu.be/AluFkuZnlyg

### DIFF
--- a/Assets/Scripts/FengGameManagerMKII.cs
+++ b/Assets/Scripts/FengGameManagerMKII.cs
@@ -159,6 +159,7 @@ public class FengGameManagerMKII : Photon.MonoBehaviour
     private bool startRacing;
     public int time = 600;
     private float timeElapse;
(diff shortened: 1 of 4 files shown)


Mi-Sad:
Well gonna remove those code smells things which is literally just 2 number not defined as const and some commented code

Mi-Sad:
Impossible to merge with how the current system is now working, is a shame i had this ready for the merge months ago and waiting just to become obsolete and totally need to be redone from 0, i won't work on this thing twice...

PR #283, opened 2020-09-18
#277 racing timer offset and string formatting

### LINKED ISSUE #277: [BUG] Racing Timers

Mi-Sad:
Solved the code smell popped up in sonarcloud on the prev pull request


Assets/Scripts/FengGameManagerMKII.cs
```diff
@@ -626,16 +626,10 @@ private void core2()
...
-                if (Gamemode.Settings.TeamMode != TeamMode.Disabled)
-                {
-                    content +=
-                        $"\n<color=#00ffff>Cyan: {cyanKills}</color><color=#ff00ff>       Magenta: {magentaKills}</color>";
-                }
-
-                this.ShowHUDInfoTopCenter(content);
-                content = Gamemode.GetGamemodeStatusTopRight((int) timeTotalServer, time);
-                this.ShowHUDInfoTopRight(content);
+                //All of this has to go as it's a double from also the previous if
```
Lithrun: So if you already discovered that this if statement does not have any impact, then please remove it.

Assets/Scripts/Gamemode/GamemodeBase.cs
```diff
@@ -204,6 +204,7 @@ public virtual void OnTitanKilled(string titanName)
             }
         }
 
+        //would be better to get float value for time and totalRoomTime instead as those are float value to begin with so this method won't have any loss of infos
```
Lithrun: Time should be managed via TimeService, but why place a comment here?

Assets/Scripts/Gamemode/RacingGamemode.cs
```diff
@@ -8,12 +8,27 @@ namespace Assets.Scripts.Gamemode
 {
     public class RacingGamemode : GamemodeBase
     {
+
+        /// <summary>
+        /// issue #277 will have to be removed once the conversion from fenggamemanager to single manager will be done
+        /// </summary>
+        public const int IntRoundTimeStatusTopScale = 10;
```
Lithrun: Use [Obsolete] instead of a summary if you intention is that this should not be used in the future

Assets/Scripts/Gamemode/RacingGamemode.cs
```diff
@@ -22,10 +37,8 @@ public override void OnGameWon()
 
             var parameters = new object[] { 0 };
             FengGameManagerMKII.instance.photonView.RPC("netGameWin", PhotonTargets.Others, parameters);
-            if (((int) FengGameManagerMKII.settings[0xf4]) == 1)
-            {
-                //this.chatRoom.addLINE("<color=#FFC000>(" + this.roundTime.ToString("F2") + ")</color> Round ended (game win).");
-            }
+
+            //there was an if on (int) FengGameManagerMKII.settings[0xf4]) == 1 and just a comment inside
```
Lithrun: Useless code, its good that you removed that, but remove your comment as well then as without that code it makes no sense

Assets/Scripts/Gamemode/RacingGamemode.cs
```diff
@@ -44,12 +57,36 @@ public override string GetVictoryMessage(float timeUntilRestart, float totalServ
...
+        /// Is just temporary as things will have to be changed and handled locally without needing to go throught FGMKII core2
+        /// </summary>
+        /// <param name="time">it require the roundtime in csec (0.1sec) so it require you to pass the time multiplied by RacingGamemode.startTimerCountdown so it will hold also info about the first dec</param>
+        /// <param name="totalRoomTime"></param>
+        /// <returns></returns>
+        public override string GetGamemodeStatusTop(int time = 0, int totalRoomTime = 0)
+        {
+            time -= RacingGamemode.IntRoundTimeStatusTopScale*RacingGamemode.StartTimerCountdown;
+            if (time > 0)
+                //if the starting time has passed it return
```
Lithrun: Your if statement already says that if time is larger than 0, so useless comments (also for the else statement)

Assets/Scripts/FengGameManagerMKII.cs
```diff
@@ -159,6 +159,7 @@ public class FengGameManagerMKII : Photon.MonoBehaviour
     private bool startRacing;
     public int time = 600;
     private float timeElapse;
+    //should be better to be able to retrive this info somehow in order to be used in other place
     private float timeTotalServer;
```
Lithrun: Rather mark this with an [Obsolete] attribute instead of a comment. Via obsolete your IDE will display a message when you are trying use the field, for a comment you'd have to look it up here.

Lithrun (changes requested):
- There are comments which do not add much of a value. Instead remove those or use the [Obsolete] attribute (with a description)

Sonarcloud will complain about that there are new [Obsolete] fields / methods used, but you can ignore those as that's something for a refactor issue.

Assets/Scripts/FengGameManagerMKII.cs
Mi-Sad: Ok i'll start reworking those than ;)

Assets/Scripts/Gamemode/GamemodeBase.cs
Mi-Sad: I'll check this timeservice thingh

Assets/Scripts/Gamemode/RacingGamemode.cs
Mi-Sad: yeah, just didn't want to put new stuff directly as obsolete :'D

Mi-Sad: Also cos it wouldn't be atm as the replacement code is not there yet

Assets/Scripts/Gamemode/RacingGamemode.cs
Mi-Sad: right

Assets/Scripts/Gamemode/RacingGamemode.cs
Mi-Sad: well just to clarify as time gets an operation done before

Assets/Scripts/FengGameManagerMKII.cs
Mi-Sad: roger

Assets/Scripts/Gamemode/GamemodeBase.cs
Lithrun: TimeService is currently only available in Issue 160 however: https://github.com/AoTTG-2/AoTTG-2/blob/%23160-titan-difficulty/Assets/Scripts/Services/TimeService.cs

Will try to do the partial merge asap as discussed so you can use that one.

Mi-Sad:
Impossible to merge with how the current system is now working, is a shame i had this ready for the merge months ago and waiting just to become obsolete and totally need to be redone from 0, i won't work on this thing twice...

Lithrun:
That's a shame indeed. Perhaps for future references, try to finish a PR when you can or allow someone else to takeover, as this one has been a draft for months.

PR #285, opened 2020-09-18
#229 auto version id retriver getting wrong branch

### LINKED ISSUE #229: [BUG] Issue versioning multiplayer conflicts
**Describe the bug**
Since the introduction of the version manager, testers are able to join lobbies of different issue versions. This is not intentional, since if players with two different versions play together, things will break. Additionally, it is also not desired that testers are able to join lobbies of an issue that a developer is working on.

**Expected behavior**
It is expected that players with the same issue number can play together, without being able to see lobbies of players with a different issue number.

Mi-Sad:
All the issue was caused by the fact that the code to get the active branch were prone to failure as it was looking for the git.exe instead of the file which stored the active branch in the repository directly, which is what it does now.
So before it just throw error if the git.exe process were not inside the AoTTG2 repository, now this condition to the installation path of git is no longer required as it check the file directly.
The system itself was already "self-sorting" the multiplayer based on the branch so i only had to make the part of code to get the current 
Cattura di schermata (1105)
branch.

Closes #229


Assets/Scripts/VersionManager.cs
```diff
@@ -29,30 +28,15 @@ private bool TryGetBranchName(ref string branchName)
...
-                branchName = process.StandardOutput.ReadLine();
-                return true;
-            }
+            string head_text = System.IO.File.ReadAllText(Directory.GetCurrentDirectory() + "\\.git\\HEAD");
+            branchName = head_text.Substring(head_text.IndexOf('#'));
+            return true;
         }
         catch (Exception e)
         {
+            Debug.Log("unable to retrive the branch name");
```
Lithrun: String should be "Unable to retrieve the branch name" (Typo in retrive)

Assets/Scripts/VersionFormatter.cs
```diff
@@ -23,15 +23,15 @@ public VersionFormatter(string issueRegex, string versionPattern)
 
     public string FormatBranchName(string branchName)
     {
-        var regex = new Regex(issueRegex);
+        Regex regex = new Regex(issueRegex);
```
Lithrun: Use var over explicitly defining a Type when it is clear what type the right hand side is

Lithrun (changes requested):
Two small maintainability comments. Will check on functionality tomorrow afternoon 👍

Lithrun (changes requested):
In terms of functionality this is not working as described per #229. I created two builds, one for issue branch 160, and one for this branch. It is expected that only players of branch #229 are able to find lobbies by making use of the "GameVersion" parameter https://doc.photonengine.com/en-us/pun/v1/getting-started/feature-overview#versioning_by_gameversion.

However, with my build of 160 I can join a lobby hosted by 229, and the other way around. This shouldn't be possible as there isn't compatibility between different versions.

Mi-Sad:
Yeah, need access to the custom server to enable that as game version in not automatically handled for self hosted server as told here from Setup And Connect tutorial from Photon 
immagine

Lithrun:
As discussed, Photon Server needs to be updated for this functionality to work. Once it does, this issue will be closed.

Mi-Sad:
Wrote a possible solution server side, but the workflow on the server components is still unclear and i can't find clear and detailed documentation. I did not test that solution yet so i have to try if it behave as expected or if everything breaks :'D

Lithrun:
Picking up this PR, since the PR was associated to #229 which was resolved thanks to a server side fix. The change to read the branch inside the .HEAD file instead of relying on `git.exe` is a cleaner approach.

Lithrun (approved):

PR #292, opened 2020-09-28
Update ServerSettingsPage.cs

### LINKED ISSUE #275: [BUG] Non-MC can change the level locally for themselves
**Describe the bug**
When a non-MC attempts to change the level, upon the next restart, this level is changed locally for them. This is undesired behavior and only the MC can determine which levels are loaded.

**To Reproduce**
1. Host a lobby a MC and join with a non-MC client
2. As non-MC, go to Game Settings > Server Settings, choose a new level and select "Change"
3. As MC, type /restart
4. The level of the MC is not changed, but the level of the non-MC is not changed.

**Expected behavior**
If a player is not the master client, then they must respect the current level that is loaded and be prevented from loading other levels.

sonicv6:


Lithrun (changes requested):
Testing Team notices an issue with this:
```
Both players start on initial level -> MC changes level and restarts -> both players on new level (as it should be) -> non-MC changes level, MC restarts -> MC remains on its level, non-MC goes back to initial level (regardless which level it chooses)
So, in short, as soon as non-MC changes the level locally it seems to be "bound" to the level it joined in on upon each restart. Would be nice if anyone else could confirm this, but I'm pretty sure that's what's happening now.
```
Please investigate and resolve it. Will change to approved once @AoTTG-2/testing-team can confirm it works as intended

sonicv6:
I'll investigate some more.

SlowDeepExhalation:
I checked the latest build, the bug Eisenlocke described seems to have been fixed.  I was unable to find any other issues with level changes not being synchronized between two clients, though I was not _completely_ exhaustive; it's possible some obscure edge cases are out there with certain maps or large numbers of players or some other oddity.  No other testers have reported problems either, so here's my go-ahead to merge it.

I don't seem to have been requested to review the PR, so this comment is all I can add.

Lithrun (approved):

PR #322, opened 2020-11-06
Hitting Code Optimization

ImBeryl:
A complete rewrite of the TriggerColliderWeapon class to continue on the necessary code refactor the old game needed.

- Replaced multiple GetComponent calls with a few TryGetComponents.
- Stored components & GameObjects as variables instead of finding them every time the code was called.
- In general, simplification of the code that should both optimize the game and allow to easily implement new features in the future.

### DIFF
--- a/Assets/Scripts/Characters/AHSSShotGunCollider.cs
+++ b/Assets/Scripts/Characters/AHSSShotGunCollider.cs
@@ -128,7 +128,7 @@ public class AHSSShotGunCollider : MonoBehaviour
                             {
                                 GameObject.Find("MainCamera").GetComponent<IN_GAME_MAIN_CAMERA>().startSnapShot2(item.transform.position, num7, null, 0.02f);
                             }
-                            item.transform.root.GetComponent<FemaleTitan>().titanGetHit(base.transform.root.gameObject.GetPhotonView().viewID, num7);
+                            item.transform.root.GetComponent<FemaleTitan>().OnNapeHitRpc2(base.transform.root.gameObject.GetPhotonView().viewID, num7, new PhotonMessageInfo());
                         }
                     }
                     else if ((item.transform.root.GetComponent<ColossalTitan>() != null) && !item.transform.root.GetComponent<ColossalTitan>().hasDie)
--- a/Assets/Scripts/Characters/AHSSShotGunCollider.cs
+++ b/Assets/Scripts/Characters/AHSSShotGunCollider.cs
@@ -140,7 +140,7 @@ public class AHSSShotGunCollider : MonoBehaviour
                         {
                             GameObject.Find("MainCamera").GetComponent<IN_GAME_MAIN_CAMERA>().startSnapShot2(item.transform.position, num8, null, 0.02f);
                         }
-                        //item.transform.root.GetComponent<ColossalTitan>().titanGetHit(base.transform.root.gameObject.GetPhotonView().viewID, num8);
+                        item.transform.root.GetComponent<ColossalTitan>().OnNapeHitRpc2(transform.root.gameObject.GetPhotonView().viewID, num8, new PhotonMessageInfo());
                     }
                     this.showCriticalHitFX(other.gameObject.transform.position);
                 }
--- a/Assets/Scripts/Characters/AHSSShotGunCollider.cs
+++ b/Assets/Scripts/Characters/AHSSShotGunCollider.cs
@@ -140,7 +140,7 @@ public class AHSSShotGunCollider : MonoBehaviour
                         {
                             GameObject.Find("MainCamera").GetComponent<IN_GAME_MAIN_CAMERA>().startSnapShot2(item.transform.position, num8, null, 0.02f);
                         }
-                        item.transform.root.GetComponent<ColossalTitan>().titanGetHit(base.transform.root.gameObject.GetPhotonView().viewID, num8);
+                        //item.transform.root.GetComponent<ColossalTitan>().titanGetHit(base.transform.root.gameObject.GetPhotonView().viewID, num8);
                     }
                     this.showCriticalHitFX(other.gameObject.transform.position);
                 }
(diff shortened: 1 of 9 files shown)


Assets/Scripts/Characters/AHSSShotGunCollider.cs
```diff
@@ -140,7 +140,7 @@ private void OnTriggerStay(Collider other)
                         {
                             GameObject.Find("MainCamera").GetComponent<IN_GAME_MAIN_CAMERA>().startSnapShot2(item.transform.position, num8, null, 0.02f);
                         }
-                        item.transform.root.GetComponent<ColossalTitan>().titanGetHit(base.transform.root.gameObject.GetPhotonView().viewID, num8);
+                        //item.transform.root.GetComponent<ColossalTitan>().titanGetHit(base.transform.root.gameObject.GetPhotonView().viewID, num8);
```
Lithrun: Why is this commented out? This seems like that the colossal titan can no longer be harmed by AHSS now, which isn't something we want. Should revert this comment

Assets/Scripts/Characters/Titan/Body/BodyPart.cs
```diff
@@ -5,6 +5,7 @@ public enum BodyPart
         None,
         Eyes,
         Head,
+        Nape,
```
Lithrun: This enum is never used

Assets/Scripts/Characters/Titan/ColossalTitan.cs
```diff
@@ -445,14 +445,14 @@ private void Start()
                 base.photonView.RPC("setSize", PhotonTargets.AllBuffered, new object[] { this.size });
                 this.lagMax = 150f + (this.size * 3f);
                 this.healthTime = 0f;
-                this.maxHealth = this.NapeArmor;
+                this.maxHealth = this.Health;
```
Lithrun: Avoid using "this" in newly added code

Assets/Scripts/Characters/Human/TriggerColliderWeapon.cs
```diff
@@ -9,360 +9,238 @@ public class TriggerColliderWeapon : MonoBehaviour
...
         {
             Transform transform = titan.transform.Find("BodyPivot/HeadPos");
-            Vector3 to = base.transform.position - transform.transform.position;
+            Vector3 to = transform.position - transform.transform.position;
             return (Vector3.Angle(-transform.transform.forward, to) < 70f);
         }
         return false;
     }
 
     public void DummyNapeHit(DummyTitan titan)
```
Lithrun, line 24: This method is unused. I think this should be somewhere within the case "titanneck"near line 109

Assets/Scripts/Characters/Human/TriggerColliderWeapon.cs
```diff
@@ -9,360 +9,238 @@ public class TriggerColliderWeapon : MonoBehaviour
...
+        titan.photonView.RPC(nameof(DummyTitan.GetHit), PhotonTargets.All, new object[] { damage });
+        manager.netShowDamage(damage);
     }
 
-    public void clearHits()
+    public void ClearHits()
     {
-        this.currentHitsII = new ArrayList();
-        this.currentHits = new ArrayList();
+        currentHitsII = new ArrayList();
```
Lithrun: Rather use .Clear() than using creating a new ArrayList

Lithrun (changes requested):
100 lines less, that feels good! A few improvements can be made, so see my comments

Assets/Scripts/Characters/Human/TriggerColliderWeapon.cs
ImBeryl: Ah, true, thought it was used somewhere else, should be fixed in the next commit.

Assets/Scripts/Characters/Human/TriggerColliderWeapon.cs
ImBeryl: Though the difference might be minimal, I remember reading that `new ArrayList()` performed better vs `.Clear()`. Fixed in the latest commit.

Assets/Scripts/Characters/AHSSShotGunCollider.cs
ImBeryl: Oh, commented it out when I was testing the changes, replaced it with the newest hit method.

Assets/Scripts/Characters/Human/TriggerColliderWeapon.cs
```diff
@@ -109,8 +92,13 @@ private void OnTriggerStay(Collider collider)
             case "titanneck":
                 if (collider.gameObject.TryGetComponent(out HitBox item) && item.transform.root.TryGetComponent(out TitanBase titanBase))
                 {
-                    if(!IsBehind(item.transform.root.gameObject)) 
-                        return;
+                    if (titanBase.transform.Find("Amarture/Core/Controller_Body/hip/spine/chest/neck/head") != null)
```
Lithrun: titanBase has a property called "TitanBody" which contains a set of bodypart references. transform.Find should be avoided as it is a very expensive method. This goes for both line 95 and 97.

I'm not too sure if the Female and Colossal titan have this component added yet. If they do not, then please add the TitanBody and assign their head object to the TitanBody Head field. Then verify if the Colossal and Female titan can still be damaged.

Lithrun (changes requested):
Final comment 👍

Assets/Scripts/Characters/Human/TriggerColliderWeapon.cs
ImBeryl: Didn't notice that one while moving code around, nice catch! Should be fixed in the most recent commit. (Tested normal, Colossal & Female titans, all were damagable)

Lithrun (approved):

PR #338, opened 2020-12-01
Events, Optimization & Achievements "prototype"

ImBeryl:
- Added some basic events that should allow better implementation of features. (Also might come in handy for mods)
- Continued refactoring of the TriggerColliderWeapon class
- Implemented a small prototype for Achievements.

### DIFF
--- a/Assets/Scripts/Achievements/AchievementManager.cs
+++ b/Assets/Scripts/Achievements/AchievementManager.cs
@@ -1,0 +1,72 @@
+﻿using Assets.Scripts.Characters.Titan;
+using Assets.Scripts.Events.Args;
+using Assets.Scripts.Services;
+using Assets.Scripts.Services.Interface;
+using UnityEngine;
+
+namespace Assets.Scripts.Achievements
+{
+    public class AchievementManager : MonoBehaviour
+    {
+        private readonly IPlayerService playerService = Service.Player;
+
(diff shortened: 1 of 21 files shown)


Assets/Scripts/Achievements/AchievementManager.cs
```diff
@@ -0,0 +1,72 @@
...
+                case TitanType.Armored:
+                case TitanType.Beast:
+                case TitanType.Colossal:
+                case TitanType.Eren:
+                case TitanType.Mindless:
+                    if (ev.Damage >= 1000)
+                    {
+                        UnlockAchievement(Achievement.DMG_1K);
+                    }
+                    else return;
```
Lithrun, line 41: Remove these "else return" statements, as you cannot obtain the 2K achievement.

Lithrun (changes requested):
For a prototype this is well made. Just remove the double "else return" and it can be approved

Assets/Scripts/Achievements/AchievementManager.cs
ImBeryl: As discussed on Discord (Leaving the comment explaining why it works for future reference) if the player deals 2K damage, it would still pass the check as it's over 1K, so it does still work! 😄

Lithrun (approved):

Issue #112, opened 2020-05-07
Weather System


Lithrun:
**Is your feature request related to a problem? Please describe.**
There will be a weather system, with Rain, Sun, Snow and Wind. Depending on the weather, these will affect your gameplay.

- [ ] Add a structure on how weather should be handled (polymorphism!) 
- [ ] Investigate how weather should exactly impact what
- [ ] Implement the weather
- [ ] Allow weather to be changed without restarting (also assure that this is synced correctly)
- [ ] Investigate how heavy the weather is on performance. If it causes frame issues, add a new graphics setting which can reduce the visual weather quality.

Prolonpo:
Likely already covered, but just leaving here for bookkeeping purposes: weather should be dynamic and evolve

From @Budeco, https://discord.com/channels/681641241125060652/683376807101071489/805449323185176646

Prolonpo:
From @ItsBlackmello, https://discord.com/channels/681641241125060652/683376807101071489/821540462421737523

> When loading into any map who have a chance to get a random type of weather. This weather could affect how you and the Titans perform on the map (i.e. rainy weather in the forest can make it slightly harder for you to grapple steadily onto trees while Titans move slower because they're walking through the mud). 

> This could be carried on through different weather patterns like 
-Fog (shortens eyesight distance)
-Snow (Makes it harder to walk on any surfaces)
-Windy (player can be pushed off axis along with their hooks)

PR #340, opened 2020-12-03
#113 time & lighting system

### LINKED ISSUE #113: Time & Lighting System
**Is your feature request related to a problem? Please describe.**
Develop a system which works on a day-night cycle. Players can configure how long a day lasts, and based on that setting, the game will automatically change its time. In the morning the sunset, and during the evening there's the sunset. During night it should be dark, so that players would have to make use of items such as torches or lanterns.

- [x] Synch Multiplayer Lobby time and lighting with all players
- [x] Add options to define the duration/interval, set time and to freeze time
- [x] Investigate on how titan AI should be affected by this

### LINKED ISSUE #526: [PROPOSAL] Upgrade to Universal Render Pipeline
This will bring many SRP features(shader graph, etc.), enhancement, and fixes the problems we are currently having with the deprecated built-in RP(flickering terrains, etc,)

1. Install Universal Render Pipeline
2. Upgrade materials in the project to use the URP shaders
3. Port custom shaders to URP

**Describe alternatives you've considered**
*Graphics*:
+ Using Amplify Shader Editor or just write one from scratch is very fine (but we still need to move to either URP or HDRP for the problems we're having)

*Fixes for the terrain flickering*:
1. Adjusting Near/Far clip of the camera.
Result: not very reliable. These methods only make the flickering less intense/noticeable. A high Near Clip value also clip out things that are too close to the camera.
1. Adjusting shadow bias.
Result: same as Near/Far clip method but make the shadows less accurate.

**Additional context** (Remove if not relevant)
This is a project I made to confirm if the flickering problem is gone.  Since this is a universal problem, any projects using built-in RP will suffer from it when terrains are far away from the scene's center. As you can see, URP have no flickering whatsoever even at 20 km away from the center. 

https://user-images.githubusercontent.com/25323231/106578035-ab06ee00-6571-11eb-9ec3-b61b150ff290.mp4

1. Determine if the enhancement is compatible with existing features
2. If the idea is denied, please provide a reason and close the issue. If the idea is accepted, consult your team to do an estimation of effort. Remove the [PROPOSAL] name and label.
3. Move the issue to the Future Milestone. On rare exceptions an issue may be moved to Release if there's sufficient upvotes.
4. Once the issue matches the current milestone, follow development procedures.

FAStre:
Build has been submitted and awaiting testing team approval. 

Closes #113 
Closes #526


FAStre:
Summary of changes:

1.  DayAndNightControl.cs : Script that controls the Time and Lighting(#113 ) system's mechanics, adapted and modified for use in AOTTG2 from a free unity store asset.
2.  TimeSwitcher.cs: Script attached to UI inputs that allows the user to edit the current time of  #113 
3.  DayLengthController.cs: Script attached to UI inputs that allows the user to edit the length of a cycle of #113 
4.  TimeSettings.cs: Script that stores the variables used for #113 , will be used along with RPC's once OnSettingsChanged event is added
5.  ToggleDayNightController.cs: Script that toggles #113 on and off.

Changes were made to all scenes, including changing lighting settings to allow for light to be influenced by colour settings rather than skybox. Additionally all ground/terrain/objects that represented a scene's 'floor' were given a standard name, "GroundTerrain" so that this component can always be grabbed by #113 and it's size used to ensure that #113 scales to fit any scene.

I do think that there are some changes that could be made in future to the time and lighting system, including features and enhancements, I'll make an issue for this soon.

FransStreicher:
Summary of changes:

1.  DayAndNightControl.cs : Script that controls the Time and Lighting(#113 ) system's mechanics
2.  TimeSwitcher.cs: Script attached to UI inputs that allows the user to edit the current time of  #113 
3.  DayLengthController.cs: Script attached to UI inputs that allows the user to edit the length of a cycle of #113 
4.  TimeSettings.cs: Script that stores the variables used for #113 , will be used along with RPC's once OnSettingsChanged event is added
5.  ToggleDayNightController.cs: Script that toggles #113 on and off.

Changes were made to all scenes, including changing lighting settings to allow for light to be influenced by colour settings rather than skybox. Additionally all ground/terrain gameobjects in maps were given a standard name, "GroundTerrain" so that this component can always be grabbed by #113 and it's size used to ensure that #113 scales to fit any scene.

I do think that there are some changes that could be made in future to the time and lighting system, including features and enhancements, I'll make an issue for this soon.

FAStre:
Returning to issue now, tester feedback yielded the following recommendations:

- Make another sunrise colour, between the twilight and dawn, add a sunrise skybox. Can be done by adding another colour "sunrise" and another material for the sunrise skybox.

- Moon sometimes is appearing as if in front of terrain when it is behind/moon rendering issues.

- Adding a sunset skybox would maybe also be good, not absolutely needed, but if it looks good it's fine to add it

- Limit a lower limit of daylength, because it is actually potentially seizure inducing at low values, could be a risk to players with epilepsy? 

- Fix colossal map lighting 

- Implement SyncSettingsRpc to replace SyncTimeRPC 

The minor bugs like the system being turned off on game restart, and when restart happens that the daylength variable(controls a daynight cycles length) is reset are minor issues I'll just make a bug issue for, with very low priority.

Lithrun (changes requested):
Request for review once you want me to review again

FAStre:
A list of bugs/work still to be done:



- [x] 1. Change IN_GAME_MAIN_CAMERA (i think) to not use the skybox component, rather have everything set from rendersettings.skybox if possible, when that's done, changing the Day and Night Control script to change skybox rendersettings instead of the main camera scripts skybox component

- [x] 2.-Multiplayer compatibility @FransStreicher 

- [x] 3.-Investigate moving to a 24h range instead of from 0-1(the original asset has 0-1) this involves not only changing the Day and Night Control script but also many other user input controls, @WattanaGaming 

- [x] 4.- Moving to gradient lighting. Wattana you seem to have experience here, I can just point to which colours we already use that would be good to include @WattanaGaming 

- [x] 5.- Sorting out the moon issue bug, if you are keen to do this @WattanaGaming 

- [x] 6. -Allow DayLength Variable and currentTIme to be set from the prefab view, this will allow map makers to simply drag paste the prefab onto a scene and set their own default lighting settings that they prefer in the scene.

- Bug fixes that are high priority: 
- [x] 7.The Colossal map has strange floor lighting when the night time of the cycle comes, not sure what is the culprit.

- [x] 8. Utgard's floor seems a bit bright compared to other maps, not as much as colossal, not sure what is the culprit.

- [x] 9. Default Skybox bug fix, where the skybox appears to move strangely(time and lighting being activated does not cause this, culprit likely due to changes to IN_GAME_MAIN_CAMERA or camera changes I think.

- [x] 10. Non MC's should not be able to influence multiplayer settings at all, even for themselves. @FransStreicher 

- [x] 11.Colossal map city floor is now extremely bright during the day.

atirutw:
> * Bug fixes that are high priority:
> * [ ]  6.The Colossal map has strange floor lighting when the night time of the cycle comes, not sure what is the culprit.
> * [ ]  7. Utgard's floor seems a bit bright compared to other maps, not as much as colossal, not sure what is the culprit.
> * [ ]  8. Default Skybox bug fix, where the skybox appears to move strangely(time and lighting being activated does not cause this, culprit likely due to changes to IN_GAME_MAIN_CAMERA or camera changes I think.
> * [ ]  9. Non MC's should not be able to influence multiplayer settings at all, even for themselves. @FransStreicher

I took a look at some of the lighting bugs described and here's the results of the initial investigation:

1. I could not confirm the strange lighting of the colossal map because I can't seem to find it in the map list when starting a game, however, the terrain floor is flickering. This is a bug with the lightmap for Unity's directional light when there's a large shadow-casting objects(for example: the terrain). Can be fixed by splitting up the terrain into smaller "tiles". Maybe we should just redo the terrain? The old one doesn't look very good anyway.
2. The direction light for the moon in Utgard is as bright as a sun(intensity is 1.4).

> * [ ]  4.- Moving to gradient lighting, adding sunrise and sunset orangish colours to. Wattana you seem to have experience here, I can just point to which colours we already use that would be good to include @WattanaGaming

You can consider this one done, btw.

FAStre:
@WattanaGaming Thanks will look at those. Colossal map can be found from the Trost map, then select Colossal Titan as an option.

FAStre:
@Lithrun Scripts/materials added for #113  can be found in Assets/Scripts/DayNightCycle and Assets/Materials/DayNightCycle. The DayNightCycle prefab is in the Resources folder. Changes to scenes are just lighting and other updates relevant to that.

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,257 @@
...
+        // Update is called once per frame
+        void Update()
+        {
+           
+
+            if (MainCamera == null)
+            {
+                try
+                {
+                    MainCamera = GameObject.Find("MainCamera").GetComponent<Camera>();
```
Lithrun: Instead of allowing an exception, you can just check if it's null. Additionally, Camera.main should return the same object (https://docs.unity3d.com/ScriptReference/Camera-main.html)

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,257 @@
...
+                }
+            }
+            frames++;
+            //MC loads settings
+            if (PhotonNetwork.isMasterClient)
+            {
+                GameSettings.Time.currentTime = currentTime;
+                GameSettings.Time.dayLength = DayLength;
+                GameSettings.Time.pause = pause;
+                Debug.Log("Current Master Client time: " + GameSettings.Time.currentTime);
```
Lithrun: Avoid logging something every frame

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,257 @@
...
+        public GameObject moon;
+        public float sunTilt = -15f;
+        [SerializeField] private TimecycleProfile timecycle = null;
+        [SerializeField] private float sunRotationOffset = 0f;
+        [Tooltip("The amount of frames to wait before doing the next lighting update")]
+        [SerializeField] private int lightingUpdateInterval = 10;
+        public Material skyBoxPROCEDURAL;
+
+        [Range(0f, 24f)] public float currentTime;
+        public float CurrentTime01 { get { return currentTime / 24; } set { currentTime = value * 24; } }
```
Lithrun: Always avoid using Numbers in your property / fields. We don't need a currentTime, currentTime1, currentTime2 ect.

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,257 @@
...
+                    targetCam = c;
+                }
+            }
+            lightIntensity = directionalLight.intensity; //what's the current intensity of the light
+            // AFAIK procedural skybox needs this to work
+            RenderSettings.sun = directionalLight;
+
+
+            //Check if default light prefab exists, and if so, disable it
+            if (GameObject.Find("LightSet"))
```
Lithrun: So we need to find this gameObject twice?

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,257 @@
...
+            
+            UpdateLight(); // Initial lighting update. Without this, the lighting will look as if it's lagging when the scene just loaded
+        }
+
+
+        public void OnPhotonPlayerConnected(PhotonPlayer newPlayer)
+        {
+            Service.Settings.SyncSettings();
+            //syncs settings after waiting a few seconds(
+            StartCoroutine(ExecuteAfterTime(5));
```
Lithrun: SyncSettings will synchronize **all** settings to everyone. Doing this when a player joins is one thing, why should do sync after 5 seconds again?

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,257 @@
...
+                // Fog
+                if (timecycle.overrideFog)
+                {
+                    RenderSettings.fogColor = timecycle.fogColor.Evaluate(CurrentTime01);
+                    RenderSettings.fogDensity = timecycle.fogColor.Evaluate(CurrentTime01).a * timecycle.maxFogDensity;
+                }
+            }
+        }
+
+        public TimeOfDay GetTimeOfDay()
```
Lithrun: If TimeOfDay.UNKNOWN means that something went wrong, it will go wrong for when CurrentTime01:
- `0`
- `>= 0.1 && <= 0.2`
- `0.5`
- `0.75`
- `1`

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,257 @@
...
+                    currentDay++; //make the day counter go up
+                }
+                if (frames == lightingUpdateInterval) { frames = 0; }
+            }
+            else
+            {
+                if (frames == lightingUpdateInterval)
+                {
+                    frames = 0;
+                    UpdateLight();
```
Lithrun: If the game is paused, then why do we need to update the light every 10 frames?

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,257 @@
...
+                return TimeOfDay.Evening;
+            }
+            else if (CurrentTime01 > 0.8f && CurrentTime01 < 1f)
+            {
+                return TimeOfDay.Night;
+            }
+            return TimeOfDay.UNKNOWN; // If this return is reached, something probably went wrong
+        }
+        
+        public enum TimeOfDay
```
Lithrun: If there is a default / unknown value, it is usually placed at the top (0)

Assets/Photon Unity Networking/Resources/PhotonServerSettings.asset
```diff
@@ -15,11 +15,11 @@ MonoBehaviour:
   AppID: 
   VoiceAppID: 
   ChatAppID: 
-  HostType: 2
+  HostType: 0
   PreferredRegion: 0
   EnabledRegions: -1
   Protocol: 0
-  ServerAddress: 51.210.5.100
```
Lithrun: Please don't change the server address and HostType

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,257 @@
...
+        public float sunTilt = -15f;
+        [SerializeField] private TimecycleProfile timecycle = null;
+        [SerializeField] private float sunRotationOffset = 0f;
+        [Tooltip("The amount of frames to wait before doing the next lighting update")]
+        [SerializeField] private int lightingUpdateInterval = 10;
+        public Material skyBoxPROCEDURAL;
+
+        [Range(0f, 24f)] public float currentTime;
+        public float CurrentTime01 { get { return currentTime / 24; } set { currentTime = value * 24; } }
+        public Slider TimeSlider  = null;
```
Lithrun: TimeSlider is unused

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,257 @@
...
+        public Camera MoonCamera = null;
+        public Camera MainCamera = null;
+        public GameObject DayNightController = null;
```
Lithrun: DayNightController is unused

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,257 @@
...
+        public int currentDay = 0;
+        public Light directionalLight;
+        public float DayLength = 300f; //default value is 300 seconds in one day
+        public bool pause { get; set; }
+        
+        [HideInInspector]
+        public float timeMultiplier = 1f; //how fast the day goes by regardless of the secondsInAFullDay var. lower values will make the days go by longer, while higher values make it go faster. This may be useful if you're siumulating seasons where daylight and night times are altered.
+        float lightIntensity; //static variable to see what the current light's insensity is in the inspector
+        public GameObject  SettingsUI = null;
```
Lithrun: timeMultiplier and SettingsUI are unused

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,257 @@
...
+       
+        // Use this for initialization
+        void Start()
+        {
+
+            pause=true;
+            RenderSettings.skybox = skyBoxPROCEDURAL;
+            Service.Settings.OnTimeSettingsChanged += Settings_OnTimeSettingsChanged;
+            MoonCamera = GetComponentInChildren<Camera>();
+            //MainCamera =
```
Lithrun: MainCamera = ?

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
Lithrun: After checking in Visual Studio, this also appears to be unused?

Assets/Scripts/DayNightCycle/DayLengthController.cs
```diff
@@ -0,0 +1,58 @@
...
+                se.AddListener(SubmitDayLength);
+                DayLengthInput.onEndEdit = se;
+            }
+        }
+
+        private void SubmitDayLength(string arg0)
+        {
+            try 
+                {
+                    float dayLength =float.Parse(arg0);
```
Lithrun: Instead of exceptions, using float.TryParse here.

Assets/Scripts/DayNightCycle/DayLengthController.cs
```diff
@@ -0,0 +1,58 @@
...
+    public class DayLengthController : MonoBehaviour
+    {
+        public Text Label;
+        public InputField DayLengthInput;
+        public Toggle ToggleDayNight;
+        DayAndNightControl dayNightCycle = null;
+
+        void Start()
+        {
+            ToggleDayNight = GameObject.Find("ToggleDayNightCycle").GetComponent<Toggle>();
```
Lithrun: ToggleDayNightCycle appears to be an UI element which is always present. You can just use a reference instead of finding it

Assets/Scripts/DayNightCycle/DayLengthController.cs
```diff
@@ -0,0 +1,58 @@
...
+            dayNightCycle = GameObject.Find("Day and Night Controller").GetComponent<DayAndNightControl>();
```
Lithrun: So does this mean that every scene requires this prefab? If so, we should document it somewhere..

Assets/Scripts/DayNightCycle/SyncTimeButtonHandler.cs
```diff
@@ -0,0 +1,43 @@
+﻿using Assets.Scripts.Services;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace Assets.Scripts.DayNightCycle
+{
+    class SyncTimeButtonHandler : MonoBehaviour
```
Lithrun: Is this class even used? I can't seem to find any references in the scenes I tested it with. (City, Outside the Walls)

Assets/Scripts/DayNightCycle/TimeSwitcher.cs
```diff
@@ -0,0 +1,84 @@
...
+        public Text Label;
+        public Slider TimeSlider;
+        public InputField TimeInput;
+        public Toggle ToggleDayNight;
+        DayAndNightControl dayNightCycle;
+
+       void Start()
+       {
+            
+            ToggleDayNight = GameObject.Find("ToggleDayNightCycle").GetComponent<Toggle>();
```
Lithrun: Once again, this ToggleDayNightCycle appears to always be present. Why use .Find instead of a reference via the inspector?

Assets/Scripts/DayNightCycle/TimeSwitcher.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+        
+        private void SubmitTime(string arg0)
+        {
+            string time = arg0;
+            if (time.Contains(":"))
+                {
+                 try 
+                    {
+                    double seconds = System.TimeSpan.Parse(time).TotalSeconds;
```
Lithrun: Once again, use .TryParse so no need to throw an exception here

Assets/Scripts/DayNightCycle/TimeSwitcher.cs
```diff
@@ -0,0 +1,84 @@
...
+             }
+            
+        }
+
+
+
+
+        public void OnDrag(PointerEventData eventData)
+        {
+            if (PhotonNetwork.isMasterClient)
```
Lithrun, line 24: This is a very expensive way of synchronizing time. Whenever this event is triggered, you are sending **all** settings to everyone in the lobby. Just by slowly dragging the bar I've sent +120 RPCs to the photon server. Settings should only be synchronized whenever the "Sync" button is pressed.

Each SyncSettings RPC is 5KB in data alone, so within just a second I've sent 600KB of data to the Photon Server, which it then also needs to send to every player.

Assets/Scripts/DayNightCycle/ToggleDayNightController.cs
```diff
@@ -0,0 +1,60 @@
...
+
+
+namespace Assets.Scripts.DayNightCycle
+{
+    public class ToggleDayNightController : MonoBehaviour
+    {
+        public Color DefaultSkyColor;
+        public Color DefaultEquatorColor;
+        public Color DefaultHorizonColor;
+        public GameObject DayNightControllerPrefab = null;
```
Lithrun: DayNightControllerPrefab, PauseDayNightButton, DayNightController, MainLight, skyBoxReset, time and seconds are all unused.

Assets/Scripts/DayNightCycle/ToggleDayNightController.cs
```diff
@@ -0,0 +1,60 @@
...
+            ToggleDayNight.isOn = false;
+            ToggleDayNight.onValueChanged.AddListener(delegate {
+                ToggleValueChanged(ToggleDayNight);
+            });
+
+        }
+
+        void ToggleValueChanged(Toggle change)
+        {
+            if (PhotonNetwork.isMasterClient)
```
Lithrun, line 32: We should only sync settings once we click the synchronize button

Assets/Scripts/Time of Day/TimecycleProfile.cs
```diff
@@ -0,0 +1,38 @@
+﻿using UnityEngine;
+
+namespace Assets.Scripts.DayNightCycle
+{
+    [CreateAssetMenu(fileName = "New Timecycle Profile", menuName = "Timecycle Profile")]
+    class TimecycleProfile : ScriptableObject
```
Lithrun: Access modifier is missing

Lithrun (changes requested):
Overall feedback:
- Name properties & fields correctly, using PascalCase. Your visual studio should give feedback about this based on our .editorconfig file. Don't use CAPSLOCK EITHER for anything
- Remove the lfs/objects files. Don't think we need those

For code feedback, see comments 
Based on testing:
Day Night Cycle > ToggleDayNightCycle gameObject has missing reference
image
The UI text is black. This makes it very difficult to see what is written on the UI 
image
Leaves are reflecting light
image

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
FAStre: We dont, will remove

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
FAStre: For some reason there seemed to be a bit of lag just as the player joined,(due to nullrefs  I think it threw off the synching of settings if the daynightcycle is running) so I added another one 5 seconds later and time seems to synch correctly. Will experiment again to see if it is no longer needed, as nullrefs were sorted out.

Assets/Scripts/DayNightCycle/TimeSwitcher.cs
FAStre: It is a tradeoff then from UX for less network use, I can put it back to just using a synch button that has to be tapped every time the slider is dragged by the user. The reason I changed it to onDrag was for UX, and I thought that once SyncSettings is expanded to only only send specific settings, so for example just time settings, that the network use would be significantly reduced. 

Alternatively we could use OnEndDrag? That way it only synchs when the slider stops being dragged, which would be equal to what would be sent once the synch button was tapped anyway.

Assets/Scripts/DayNightCycle/ToggleDayNightController.cs
FAStre: Yes will remove these.

Assets/Scripts/DayNightCycle/ToggleDayNightController.cs
FAStre: This is again a UX decision, the user would have to anyway tap a synch button for this rpc to take effect, and this one only synchs once a toggle, so it would be no more RPC's being sent than it would with a synch button being pressed after changing the toggle value.

Assets/Scripts/Time of Day/TimecycleProfile.cs
FAStre: @WattanaGaming

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
FAStre: Will do, @WattanaGaming would you like to tackle this as it is your implementation? I can also take it if you want, either way is fine with me.

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
FAStre: It seems this is a leftover from debugging, will remove it.

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
FAStre: Will change font and remove the unused prefab reference. @WattanaGaming would you be able to look at the leaves reflecting? If it is something to do with the leaf material/texture then I think it might be another issue not related to time and lighting, but if we can fix it with lighting settings then we should.

Assets/Scripts/DayNightCycle/ToggleDayNightController.cs
FAStre: Well even ignoring UX, sending it on ToggleValueChanged is the same as not having it and tapping the synch button(which some users may tap more than once) after changing the toggle setting. By having it here it makes it easier to use and sends the same/less  RPC data anyway.

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
atirutw: `currentTime` is what we use for the 24 hour format and `CurrentTime01` is for 0-1 format as to not have `currentTime / 24` everywhere. Hopefully this is a good excuse lol.

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
atirutw: @FransStreicher fixed 👍

Assets/Scripts/Time of Day/TimecycleProfile.cs
atirutw: Whoops. Will fix it right away

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
FAStre: Strange, initially not catching a nullref caused issues for the non-mc, but now it doesn't anymore, will push the changes without the try catch here.

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
FAStre: @WattanaGaming  I think you added the unknown

Assets/Scripts/DayNightCycle/DayLengthController.cs
FAStre: Yes, I left a comment on #365 that once #113 is done, that it should be included on every scene, and added to the mapping prefabs folder for mappers.

Assets/Scripts/DayNightCycle/SyncTimeButtonHandler.cs
FAStre: It used to be the syncbutton script, will remove it.

Assets/Scripts/DayNightCycle/TimeSwitcher.cs
FAStre: Will do

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
FAStre: Ok checked on a rebuild and once again getting a nullref here, only for non-MC users though. For some reason the Main Camera is not loaded in time, but only on multiplayer and for non-MC's, so this is a workaround for that, and is necessary to avoid nullref errors.

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
atirutw: @FransStreicher wdym? I only modified it to use enum

FAStre: @WattanaGaming https://github.com/AoTTG-2/AoTTG-2/commit/97b2e408a1d91d1659ca7014c6ebff0192d2efcd here you changed it to enum and added the 'unknown' line of code, I'm wondering whether we need it? I can add a = sign to the > signs in the if statements so that time will always be one of the other values(morning,afternoon,evening,midnight), in that case, I don't know if we will need unknown as after adding equals signs in the if statements, any currentTime will always have a value 0<=x<=1 so perhaps no need for an unknown returning.

FransStreicher: Seems I was overlooking that it's required, will just move the code of line to the top as Lithrun mentioned defaults are usually at the top

FAStre: I think the line of code is needed and makes sense to be where it is, will just update some of the if conditions.

FAStre:
Ok it seems like there's a bit more to do.

- [x] Moving multiplayer compatibility to use OnJoinedRoom has made multiplayer settings not synch anymore, whereas before it was working fine. Using OnJoinedRoom is however a cleaner solution so it should be investigated why this isnt working with time settings @FransStreicher 


- [x] Re-add the SynchTime button, instead of people using enter on the input fields, rather keep it as a button to submit all time input fields. @FransStreicher -After consulting with testers, for now enter for submit will stay, when UI is reworked project wide it may change to a button or not.

- [x] Fixing the shadow flickering on some maps. @WattanaGaming you mentioned looking into this. -will be solved by #526

Toorah:
> *  Integrate sky, moon and stars rendering to just one camera, the main camera(and remove MoonCamera), should improve performance @Toorah you seemed to be keen to work on this.

Yes, I can do that maybe tomorrow

> *  Implementing basic clouds (optional) @Toorah you mentioned working on this.

I tested something yes, though I think it will need more time, and I don't know if it's rather linked to the Weather System

FAStre:
> > * Integrate sky, moon and stars rendering to just one camera, the main camera(and remove MoonCamera), should improve performance @Toorah you seemed to be keen to work on this.
> 
> Yes, I can do that maybe tomorrow
> 
> > * Implementing basic clouds (optional) @Toorah you mentioned working on this.
> 
> I tested something yes, though I think it will need more time, and I don't know if it's rather linked to the Weather System

Cool, if clouds are better kept in #112 , then I dont think we have to add it now, the stars that have been recently added already improve the sky enough for a preview release/devblog.

atirutw:
> * [ ]  Integrate sky, moon and stars rendering to just one camera, the main camera(and remove MoonCamera), should improve performance @Toorah you seemed to be keen to work on this.(optional)

This is impossible without the moon getting obscured by fog.

Moon behind everything = gets obscured by fog,
Moon close up and small = renders in front of everything

The only way (that I know) to make the moon render behind everything without getting obscured by the fogs is to use a second camera.

atirutw:
Some quality of life stuff I wanna implement. They're optional but will probably help feed my ego

- [x] Make `DayAndNightControl.cs` preview changes in editor using `OnValidate()`.
- [x] Try to make the codes look even better and more readable.

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,247 @@
...
+        public Slider TimeSlider  = null;
+        public Camera MoonCamera = null;
+        public Camera MainCamera = null;
+        public int currentDay = 0;
+        public Light directionalLight;
+        public float DayLength = 300f; //default value is 300 seconds in one day
+        public bool pause { get; set; }
+        
+        [HideInInspector]
+        float lightIntensity; //static variable to see what the current light's insensity is in the inspector
```
WattanaGaming, line 31: > "to see what the current light's insensity is in the inspector"

> Uses a private float that have the `HideInInspector` attribute

This feel so wrong on many levels

atirutw: @FransStreicher are you sure you're doing this one right? You want it to be visible in the editor but you used a private float and gave it the `HideInInspector` attribute on top of that.

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
FAStre: Because the user can change the cycle even if the game is paused.

Assets/Editor/PrefabReplacer.cs
```diff
@@ -0,0 +1,43 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEditor;
+using UnityEngine;
+
+namespace Assets.Editor.Prefabs
+{
+    public class PrefabReplacer : EditorWindow
```
Lithrun: PrefabReplacer seems to have the same functionality as the "GameObjectToPrefabConverter", please merge the two together instead of having two (near) identical tools

Assets/Photon Unity Networking/Resources/PhotonServerSettings.asset
```diff
@@ -15,7 +15,7 @@ MonoBehaviour:
   AppID: 
   VoiceAppID: 
   ChatAppID: 
-  HostType: 2
+  HostType: 0
```
Lithrun: Please don't change the host type

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,246 @@
...
+    public class DayAndNightControl : MonoBehaviour
+    {
+        public GameObject moon;
+        public ReflectionProbe reflectionProbe;
+        public float sunTilt = -15f;
+        [SerializeField] private TimecycleProfile timecycle = null;
+        [SerializeField] private float sunRotationOffset = 0f;
+        [Tooltip("The amount of frames to wait before doing the next lighting update")]
+        [SerializeField] private int lightingUpdateInterval = 10;
+        public Material skyBoxPROCEDURAL;
```
Lithrun: Apply PascalCase

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,246 @@
...
+using UnityEngine;
+using UnityEngine.UI;
+using UnityEngine.Animations;
+using UnityEditor;
+using Assets.Scripts.Services;
+using Assets.Scripts.Settings;
+
+namespace Assets.Scripts.DayNightCycle
+{
+    public class DayAndNightControl : MonoBehaviour
```
Lithrun, line 15: Public properties & fields are always PascalCase

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,246 @@
...
+        public Slider TimeSlider  = null;
+        public Camera MoonCamera = null;
+        public Camera MainCamera = null;
+        public int currentDay = 0;
+        public Light directionalLight;
+        public float DayLength = 300f; //default value is 300 seconds in one day
+        public bool pause { get; set; }
+        public float lightIntensity; //static variable to see what the current light's insensity is in the inspector
+
+        Camera targetCam;
```
Lithrun: targetCamera appears to be unused, also missing its access modifier

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
Lithrun: TimeSlider is still unused

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,246 @@
...
+        private int frames;
+       
+        // Use this for initialization
+        void Start()
```
Lithrun, line 38: Always be explicit when it comes to access modifiers

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
Lithrun: Also remove the default unity comments

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,246 @@
...
+                {
+                    targetCam = c;
+                }
+            }
+            lightIntensity = directionalLight.intensity; // What's the current intensity of the sunlight
+
+            UpdateLightingSettings();
+            UpdateLight(); // Initial lighting update.
+
+            IEnumerator SetupCameraConstraintAndReflectionProbe()
```
Lithrun, line 57: Local Functions are always at the bottom of the method

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,246 @@
...
+        public Material skyBoxPROCEDURAL;
+
+        [Range(0f, 24f)] public float currentTime;
+        public float CurrentTime01 => currentTime / 24;
+        public Slider TimeSlider  = null;
+        public Camera MoonCamera = null;
+        public Camera MainCamera = null;
+        public int currentDay = 0;
+        public Light directionalLight;
+        public float DayLength = 300f; //default value is 300 seconds in one day
```
Lithrun: DayLength = 300f. Default value is 300, I can see that. Rather add a /// summary instead of a comment, or remove it

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,246 @@
...
+            {
+                // 300s since MC updated this
+                var diff = (float) (DateTime.UtcNow - settings.LastModified).TotalSeconds;
+                // 5 += ( 300 / 300) * 24 => 29 % 24 => 5
+                currentTime += (diff / DayLength) * 24;
+                //If time passed will put the currentTime over 24, do maths to correct for this and give an accurate
+                //time according to a 24h time range
+                if (currentTime > 24)
+                {
+                    currentTime = currentTime - (24 * (int)Math.Floor(currentTime/24));
```
Lithrun: Could use a `currentTime -=` instead of `currentTime = currentTime -`

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,246 @@
...
+            }
+            else
+            {
+                MoonCamera.fieldOfView = MainCamera.fieldOfView;
+            }
+
+            if (!pause)
+            {
+                currentTime += (Time.deltaTime / DayLength) * 24;
+                if (CurrentTime01 >= 1) // If CurrentTime >= 24 hours
```
Lithrun: CurrentTime01 is still a vague property

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,246 @@
...
+            skyBoxPROCEDURAL.SetFloat("_Angle", -CurrentTime01 * 360f);
+        }
+
+        void UpdateLightingSettings()
+        {
+            RenderSettings.skybox = skyBoxPROCEDURAL;
+            RenderSettings.sun = directionalLight; // Procedural skybox needs this to work
+            RenderSettings.fog = true;
+
+            if (timecycle)
```
Lithrun: if statement can be inverted to make the code more readable => `if (!timeCycle) return;`

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,246 @@
...
+        void UpdateLight()
+        {
+            Quaternion tilt = Quaternion.AngleAxis(sunTilt, Vector3.forward);
+            Quaternion rot = Quaternion.AngleAxis((CurrentTime01 * 360) - 90, Vector3.right);
+
+            directionalLight.transform.rotation = tilt * rot; // Yes axial tilt
+            directionalLight.transform.Rotate(Vector3.up, sunRotationOffset - 90, Space.World);
+            moon.transform.forward = -directionalLight.transform.forward;
+
+            if (timecycle)
```
Lithrun: #InvertThatIfStatement

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,246 @@
...
+        [SerializeField] private TimecycleProfile timecycle = null;
+        [SerializeField] private float sunRotationOffset = 0f;
+        [Tooltip("The amount of frames to wait before doing the next lighting update")]
+        [SerializeField] private int lightingUpdateInterval = 10;
+        public Material skyBoxPROCEDURAL;
+
+        [Range(0f, 24f)] public float currentTime;
+        public float CurrentTime01 => currentTime / 24;
+        public Slider TimeSlider  = null;
+        public Camera MoonCamera = null;
```
Lithrun, line 28: Shouldn't the MoonCamera be disabled if you are within a time of day where the moon is not present? In your Update method, you are changing the FOV of the moon every frame, yet what is the point if the moon is not visible?

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,246 @@
...
+                // Fog
+                if (timecycle.overrideFog)
+                {
+                    RenderSettings.fogColor = timecycle.fogColor.Evaluate(CurrentTime01);
+                    RenderSettings.fogDensity = timecycle.fogColor.Evaluate(CurrentTime01).a * timecycle.maxFogDensity;
+                }
+            }
+        }
+
+        void OnValidate()
```
Lithrun, line 232: "This function is called when the script is loaded or a value is changed in the Inspector (Called in the editor only)."

So, since this can only be called by the editor, where is there an extra #if UNITY_EDITOR region?

Assets/Scripts/DayNightCycle/DayLengthController.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+    public class DayLengthController : MonoBehaviour
+    {
+        public Text Label;
+        public InputField DayLengthInput;
+        DayAndNightControl dayNightCycle;
+
+        void Start()
+        {
+            dayNightCycle = GameObject.Find("Day and Night Controller").GetComponent<DayAndNightControl>();
```
Lithrun: Let's not use hardcoded strings to find something. Where are the references?

Assets/Scripts/DayNightCycle/DayLengthController.cs
```diff
@@ -0,0 +1,56 @@
+﻿using UnityEngine;
+using UnityEngine.UI;
+using Assets.Scripts.Services;
+using Assets.Scripts.Settings;
+using UnityEngine.SceneManagement;
+namespace Assets.Scripts.DayNightCycle
+{
+    public class DayLengthController : MonoBehaviour
```
Lithrun, line 13: Access modifiers are great

Assets/Scripts/DayNightCycle/DayLengthController.cs
```diff
@@ -0,0 +1,56 @@
...
+        void Start()
+        {
+            dayNightCycle = GameObject.Find("Day and Night Controller").GetComponent<DayAndNightControl>();
+        }
+
+        void Update()
+        {
+            if (PhotonNetwork.isMasterClient)
+            {
+                var se = new InputField.SubmitEvent();
```
Lithrun: We don't need to do this every frame. Could just be moved to Start / Awake

Assets/Scripts/DayNightCycle/DayLengthController.cs
```diff
@@ -0,0 +1,56 @@
...
+                        dayNightCycle.DayLength = dayLength;
+                        GameSettings.Time.dayLength = dayLength;
+                        Service.Settings.SyncSettings();
+                    }      
+        }
+
+        //grabbing the local scene's DayAndNightControl script, and adding a scene changed listener
+        void OnEnable()
+        {
+            dayNightCycle = GameObject.Find("Day and Night Controller").GetComponent<DayAndNightControl>();
```
Lithrun, line 40: References instead of .Find

Assets/Scripts/DayNightCycle/DayLengthController.cs
```diff
@@ -0,0 +1,56 @@
...
+            SceneManager.sceneLoaded += OnSceneLoaded;
```
Lithrun, line 41: If we subscribe to an event OnEnable, then we also need to UnSubscribe OnDisable

Assets/Scripts/DayNightCycle/TimeSwitcher.cs
```diff
@@ -0,0 +1,86 @@
...
+
+
+
+
+        //grabbing the local scene's DayAndNightControl script, and adding a scene changed listener
+        void OnEnable()
+        {
+            dayNightCycle = GameObject.Find("Day and Night Controller").GetComponent<DayAndNightControl>();
+            TimeSlider.value = dayNightCycle.CurrentTime01;
+            SceneManager.sceneLoaded += OnSceneLoaded;
```
Lithrun, line 74: Don't forget to UnSubscribe, otherwise we will keep subscribing whenever this is enabled

Assets/Scripts/DayNightCycle/ToggleDayNightController.cs
```diff
@@ -0,0 +1,51 @@
...
+                dayNightCycle.pause = !change.isOn;
+                GameSettings.Time.pause = !change.isOn;
+                Service.Settings.SyncSettings();
+            }
+
+        }
+        void OnEnable()
+        {
+            dayNightCycle = GameObject.Find("Day and Night Controller").GetComponent<DayAndNightControl>();
+            SceneManager.sceneLoaded += OnSceneLoaded;
```
Lithrun, line 43: Don't forget to UnSubscribe and click that dislike button. Also checkout my Patreon to get money

Assets/Scripts/Settings/TimeSettings.cs
```diff
@@ -0,0 +1,34 @@
...
+        public TimeSettings(Difficulty difficulty)
+        {
+            switch (difficulty)
+            {
+                case Difficulty.Easy:
+                case Difficulty.Normal:
+                case Difficulty.Hard:
+                case Difficulty.Abnormal:
+                case Difficulty.Realism:
+                    dayLength = 1000;
```
Lithrun: All settings should be assigned a value, otherwise you will get NullReferenceExceptions

Assets/Scripts/Settings/TimeSettings.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Assets.Scripts.Settings
+{
+    public class TimeSettings
+    {
+       
+        public float? currentTime { get; set; }
+        public float? dayLength { get; set; }
+        public bool? pause { get; set; }
+        public DateTime LastModified { get; set; }
```
Lithrun: Must be of type DateTime?

Assets/Scripts/UI/InGame/ServerSettingsPage.cs
```diff
@@ -1,4 +1,5 @@
-﻿using Assets.Scripts.Settings.Gamemodes;
+﻿using Assets.Scripts.Services;
+using Assets.Scripts.Settings.Gamemodes;
 using System.Collections.Generic;
 using System.Linq;
 using UnityEngine;
```
Lithrun: Lots of random newlines, seems unnecessary

ProjectSettings/ProjectVersion.txt
```diff
@@ -1,2 +1,2 @@
-m_EditorVersion: 2020.1.17f1
-m_EditorVersionWithRevision: 2020.1.17f1 (9957aee8edc2)
+m_EditorVersion: 2020.2.1f1
```
Lithrun: Should keep in mind that 2020.2.1f1 is not possible until #373

Lithrun (commented):
Alright Pull Request review time... 166 files, fun time~

For individual code comments, see below. Overall:

**Quality Settings - Shadows**
With the introduction of this issue, the nights will look a lot darker due to the shadows, which is great. Yet, is it intentional that shadows can be disabled on the lowest quality settings? If so, playing on a night mode with disabling shadows gives them a significant advantage:

Low Settings
image

Ultra settings
image

**Performance**
Haven't been able to test in depth with profiler, but will see what testing team reports first.

**Other**
- Assets/New Terrain 1.asset what does this terrain asset do? Leftover? If so, delete
- Assure that there are always access modifiers
- What are these files inside "hooks/" ?

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
FAStre: I can put an if statement for the times when the moon is definitely below terrain and unobservable, then the FOV is only updated when the moon is visible

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
FAStre: @WattanaGaming  I think you added this would you like to take it up?

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
FAStre: @WattanaGaming Your addition as well, please check it out if you have time to.

Assets/Scripts/DayNightCycle/DayLengthController.cs
FAStre: I don't know if I can reference the Day and Night Controller because it is added to every scene, so basically the only way to find it is when the UI component is enabled in a scene, to then find the local Day and Night Controller gameobject.

Also to note DayAndNightControl.cs itself is about 0.3% of CPU ms usage so doesn't significantly affect performance. Lighting and Rendering is by far the larger culprit of performance drops thus far I think(from unity profiler)

Assets/Scripts/DayNightCycle/DayLengthController.cs
FAStre: I don't know if I can reference the Day and Night Controller because it is added to every scene, so basically the only way to find it is when the UI component is enabled in a scene, to then find the local Day and Night Controller game object.

Also to note DayAndNightControl.cs itself is about 0.3% of CPU ms usage so doesn't significantly affect performance. Lighting and Rendering is by far the larger culprit of performance drops thus far I think(from unity profiler)

Assets/Editor/PrefabReplacer.cs
atirutw: @Toorah you made it so.... *shrugs*

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
atirutw: That was my habit of putting local functions right above where they're used. Will fix

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
atirutw: > I can put an if statement for the times when the moon is definitely below terrain and unobservable, then the FOV is only updated when the moon is visible

Does this not also effect the sun?

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
FAStre: Does even the internal unity monobehaviours of a class need access modifiers? The class itself has one so idk if void Start() and void Update need one too, havnt seen any other ones in the project with access modifiers.

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
FAStre: It does seem so from some quick testing, so i dont think we can disable this.

Assets/Scripts/Gamemode/GamemodeBase.cs
```diff
@@ -172,6 +172,9 @@ public virtual TitanConfiguration GetTitanConfiguration(MindlessTitanType type)
 
         public virtual void OnRestart()
         {
+            Debug.Log("CLEARING THAT CASH");
```
Lithrun: Remove old comment

Assets/Scripts/Gamemode/RacingGamemode.cs
```diff
@@ -99,9 +102,20 @@ private void RequestStatus(PhotonMessageInfo info)
         private void RacingStartRpc(PhotonMessageInfo info)
         {
             if (!info.sender.IsMasterClient) return;
+            Debug.Log(nameof(RacingStartRpc));
```
Lithrun: Don't need log comments like those

Assets/Scripts/Gamemode/RacingGamemode.cs
```diff
@@ -99,9 +102,20 @@ private void RequestStatus(PhotonMessageInfo info)
...
+            if (IsLoaded)
+            {
+                yield return new WaitForSeconds(0.05f);
+            }
 
-            StartBarriers.ForEach(x => x.gameObject.SetActive(false));
-            UiService.ResetMessage(LabelPosition.Center);
+            FindObjectsOfType<RacingStartBarrier>().ToList()
+                .ForEach(x => x.gameObject.SetActive(false));
+            Debug.Log("Running Couroutine");
```
Lithrun: nor this

Assets/Scripts/Settings/TimeSettings.cs
```diff
@@ -0,0 +1,34 @@
...
+using System;
+
+namespace Assets.Scripts.Settings
+{
+    public class TimeSettings
+    {
+       
+        public float? currentTime { get; set; }
+        public float? dayLength { get; set; }
+        public bool? pause { get; set; }
```
Lithrun: Current time and pause do not have a default setting, thus there may be a null reference exception.

Assets/Scripts/Settings/TimeSettings.cs
```diff
@@ -0,0 +1,34 @@
+﻿
+using Assets.Scripts.Gamemode;
+using System;
+
+namespace Assets.Scripts.Settings
+{
+    public class TimeSettings
+    {
+       
+        public float? currentTime { get; set; }
```
Lithrun: public keyword is always **P**ascal**C**ase. This isn't Java >.>

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,248 @@
...
+        private int frames;
+       
+        void Start()
+        {
+            Pause=true;
+            Service.Settings.OnTimeSettingsChanged += Settings_OnTimeSettingsChanged;
+            //MC loads settings on start because these values need to have a default.
+            if (PhotonNetwork.isMasterClient)
+            {
+                GameSettings.Time.currentTime = CurrentTime;
```
Lithrun: These are either supplied by the Difficult or the users custom settings. Because of the current implementation: the day night cycle cannot be paused, start with a specific time or specify what the length of a day should be by a global, gamemode or map setting. Remove such limitation to the fancy system.

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -0,0 +1,248 @@
...
+            StartCoroutine(SetupCameraConstraintAndReflectionProbe());
+            IEnumerator SetupCameraConstraintAndReflectionProbe()
+            {
+                // Set up rotation constraint for the moon camera.
+                RotationConstraint constraint = MoonCamera.gameObject.GetComponent<RotationConstraint>();
+
+                ConstraintSource source = new ConstraintSource();
+                Debug.Log("Waiting for Main Camera...");
+                yield return new WaitUntil(() => MainCamera != null);
+                Debug.Log("The main camera has arriveth");
```
Lithrun: These logs don't appear that useful

Lithrun (changes requested):
Final comments. Good luck

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
FAStre: Sounds good, at the moment it is sort of relegated to map setting, except via the prefab, so I can see if there is a place to set map settings and make it so the prefab allows the editing of those settings to for ease of use for mappers.

Assets/Scripts/Settings/TimeSettings.cs
FAStre: Currently set from the prefabs variables, of which maybe map settings are better, will check that.

Assets/Scripts/DayNightCycle/DayAndNightControl.cs
Lithrun: Should still be via the TimeSetting, as otherwise it won't work well for custom maps, nor for my "Lithruns Hardcore Capture Gamemode Night 24/7 Edition" difficulty setting.

FAStre: I see yeah, we can discuss the implementation over VC sometime.

Lithrun (approved):
Alright, time to merge this one 😄

PR #377, opened 2020-12-21
Update InputManager.cs

sidsayshmm:
Fixed NREs due to PlayerPrefs not being updated/loaded properly in InputManager.cs


Lithrun (approved):
And another bug down! Just many more to go 👍

PR #378, opened 2020-12-21
#190 drp

### LINKED ISSUE #190: Discord Rich Presence (DRP) integration
**Is your feature request related to a problem? Please describe.**
Discord is one of the largest social platforms for gamers. It has a feature named Rich Presence, which allows users to see what games their friends are playing, but also show some additional information. An example of this is 
image

A Discord Rich Presence integration which displays various information (See below) of what the user is currently playing.

- [x] Region
- [x] Lobby name
- [x] Lobby gamemode & level information (Example: The Forest - Waves, Utgard Castle - Capture ect.)
- [x] Current / Max Players
- [x] An option to join the lobby that the discord user is playing. If it is public, then the user can automatically join. If it's hosted by Custom or contains a password, then an "Ask To Join" request has to be sent.

sidsayshmm:
Sets up rich presence, with different discord status in Menu, Single player, and multiplayer mode. 
Displays an "Ask To Join" button if playing in a multiplayer game, which sends a request to the player as a discord DM, which on accepting, gives out a "Join" option in the discord DM. 

On accepting, launches the game if its not open (requires the game to be started at-least once), and then joins the room displayed in the discord status.

Changes : 
1) Deletes the maxPlayers property from FengGameManager.
2) Creation of PhotonService, to reduce code duplication, (also includes changes to MainMenu.cs and Lobby.cs) 
3) Creation of DiscordService to manage the discord rich presence.

Closes #190


Assets/Scripts/Services/DiscordService.cs
```diff
@@ -0,0 +1,199 @@
...
+                        : $"Failed to Update Multi-player party.");
+                });
+        }
+
+        #region Helper Methods
+
+        private static string GetApplicationPath()
+        {
+            string path = System.Environment.CurrentDirectory;
+            path += "\\";
```
Lithrun: This `path += "\\";` may not work on OSX. Windows use a `\` as file separator, whereas Mac uses a `/` as file separator. Rather use `Path.DirectorySeparatorChar` https://docs.unity3d.com/ScriptReference/Path.DirectorySeparatorChar.html

Assets/Scripts/Services/DiscordService.cs
```diff
@@ -0,0 +1,199 @@
...
+            switch (Application.platform)
+            {
+                case RuntimePlatform.WindowsPlayer:
+                    path += "AoTTG 2.exe";
```
Lithrun: Perhaps instead of hard coding "AoTTG 2" here, we should use `Application.companyName`. Whenever Unity does a build, it will name the .exe / .app after this value. (https://docs.unity3d.com/ScriptReference/Application-companyName.html)

Assets/Scripts/Services/Interface/IDiscordService.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Assets.Scripts.Services.Interface
+{
+    public interface IDiscordService
+    {
+        void JoinViaDiscord(string roomID);
```
Lithrun: JoinViaDiscord is only used within DiscordService. Could be made private if we don't intend to expose this method

Lithrun (changes requested):
Just some minor code improvements, and one potential bug.
Another issue I noticed:
- [ ] If you are running the game already, and open discord afterwards, then it will not display the Rich Presence.

Issue #351, opened 2020-12-09
[BUG] No FPS cap on Main Menu


Acros5542:
**Describe the bug**
The main menu currently is using a very high amount of system resources because it does not have an FPS cap set in place.

**To Reproduce**
1. Launch AoTTG2 and sit on the main menu
2. Using some sort of FPS tracker (shadowplay has one built in you can turn on), check the FPS the main menu is running at
3. Additionally you can use task manager to see it's using a very high amount of system resources

**Expected behavior**
Main menu should have an FPS cap so simply leaving the game up and idling isn't overly stressing your system.

**Screenshots**
(on a 2080TI graphics card)
https://cdn.discordapp.com/attachments/709670139276165152/783941096870248478/Screenshot_161.png
https://cdn.discordapp.com/attachments/709670139276165152/783946866227544074/Screenshot_277.png

**Contributor instructions**
1. Try to recreate the bug via the reported reproduction steps. If the bug can be reproduced, remove the investigation label. If the bug cannot be reproduced, please add a comment, attach "On Hold" label and inform the submitter that the bug cannot be reproduced. If there is no reply after 2 weeks, the issue can be closed.
2. Follow the default development team procedures (Create a new branch, apply your fix, create a pull request, submit new builds for testing team in dev squad)
3. Once testing team has confirmed the issue is resolved and the pull request is accepted, the issue can be closed.

Lithrun:
Resolving this is fairly simply as well. When the main menu is opened, set the FPS limit to the refresh rate (Hz) of the monitor. https://docs.unity3d.com/ScriptReference/Application-targetFrameRate.html

Once the player goes in-game, the FPS limit should respect the users settings.

PR #379, opened 2020-12-21
#351 - FPS Cap on main menu

ronitrocket:


Assets/Scripts/FPSCapper.cs
```diff
@@ -0,0 +1,34 @@
...
+
+
+        // Update is called once per frame
+        void LateUpdate()
+        {
+            float curTime = Time.realtimeSinceStartup;
+            float timeTaken = (curTime - oldTime);
+            if (timeTaken < theDeltaTime)
+            {
+                Thread.Sleep((int) (1000 * (theDeltaTime - timeTaken)));
```
Lithrun, line 27: Thread.Sleep should never be used in Synchronous code as it blocks the entire thread. If you want to wait a specific amount of time, you should rather use Coroutines (https://docs.unity3d.com/Manual/Coroutines.html) if you really have to wait.

Although, what does this code actually do? It sets oldTime and theDeltaTime variables, but those are never used in the application

Assets/Photon Unity Networking/Resources/PhotonServerSettings.asset
```diff
@@ -15,14 +15,14 @@ MonoBehaviour:
   AppID: 
   VoiceAppID: 
   ChatAppID: 
-  HostType: 2
```
Lithrun, line 18: Should undo these changed to the PhotonServerSettings as merging these into development will break online functionality.

Assets/FPSCapper.prefab
```diff
@@ -0,0 +1,45 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!1 &9169474616986226046
```
Lithrun: Besides that we don't need this prefab, also avoid placing prefabs inside the root folder

Lithrun (changes requested):
I pulled the branch and the code you submitted does not fix the issue. Are you sure you committed the right files? Since based on the code you submitted, I can't seem to find which part of your code is supposed to fix the issue.

Also see the other comments.

Lithrun:
Issue #351 is closed, so this PR can be closed as well.

PR #384, opened 2020-12-22
#367 kill dmg not clearing from the screen

### LINKED ISSUE #367: [BUG] Kill Dmg not clearing from the screen
**Describe the bug**
By leaving a single or multiplayer game while the yellow damage for a kill is still on screen (not just in the kill feed but the larger number that displays immediately after killing a titan), upon joining another lobby this damage number will be permanently stuck on screen. If you get another kill while in this state, this new number will replace the old one, but then it will not clear successfully itself. Closing the client and relaunching fixes this, /restart does not. (Also sorry for no video, since moving back home I don't have the upload speed to reasonably upload videos anymore, but I can go back and upload all needed footage relatively quickly in about a month from now - if footage is needed for anything let me know and I can get it uploaded in about 24-48 hours from then depending on how long of a clip.)

**To Reproduce**
1. Create a single or multiplayer lobby
2. Kill a titan and immediately leave the lobby before the dmg number leaves the screen
3. Create a new single or multiplayer lobby

**Expected behavior**
The kill damage should correctly clear from the screen and not persist when the game is restarted or left.

**Log file**
https://www.dropbox.com/s/vth5pnfl7h3x4vj/Dmg%20number%20bug.txt?dl=0

**Additional context**
This works on both Alpha 0.1 and Alpha 0.2

**Contributor instructions**

8veStory:
Applied fix by resetting kill info and damage UI elements upon level load. Change seems to work as expected. Closes #367.

### DIFF
--- a/Assets/Scripts/FengGameManagerMKII.cs
+++ b/Assets/Scripts/FengGameManagerMKII.cs
@@ -2827,9 +2827,9 @@ namespace Assets.Scripts
         [PunRPC]
         private void updateKillInfo(bool t1, string killer, bool t2, string victim, int dmg)
         {
-            GameObject killFeed = GameObject.Find("KillFeed");
-            GameObject newKillInfo = (GameObject) UnityEngine.Object.Instantiate(Resources.Load("UI/KillInfo"));
-            foreach (GameObject killInfo in killInfoGO)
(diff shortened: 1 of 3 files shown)


Lithrun (approved):
And another bug down, you're doing awesome 👍

PR #427, opened 2020-12-30
Fixed player titan sliding around on death

### LINKED ISSUE #389: [BUG] Player titans moving on death
**Describe the bug**
If you suicide while moving as a player titan your model slides across the floor of the map until it despawns.

**To Reproduce**
1. Create a multiplayer lobby on forest
2. Spawn as a titan
3. Hold W
4. Suicide while still holding W

**Expected behavior**
Your model should fall and remain motionless after suiciding.

**Additional context**
I'm guessing this is because your speed isn't reset to 0 after suiciding, so your model continues to slide at the speed you were moving before suiciding.

**Contributor instructions**

mash8:
Closes #389

### DIFF
--- a/Assets/Scripts/Characters/Titan/MindlessTitan.cs
+++ b/Assets/Scripts/Characters/Titan/MindlessTitan.cs
@@ -521,6 +521,7 @@ namespace Assets.Scripts.Characters.Titan
         private bool HasDieSteam { get; set; }
         protected void Dead()
         {
+            Rigidbody.velocity = new Vector3(0, Rigidbody.velocity.y, 0);
             if (!Animation.IsPlaying(AnimationDeath))
             {
                 CrossFade(AnimationDeath, 0.05f);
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Characters/Titan/MindlessTitan.cs
```diff
@@ -521,6 +521,7 @@ private void Explode()
         private bool HasDieSteam { get; set; }
         protected void Dead()
         {
+            Rigidbody.velocity = new Vector3(0, Rigidbody.velocity.y, 0);
```
Lithrun: Don't think we need this line of code, as that will instantly kill any existing speed which a titan has. The check to see if the titan is alive in PlayerTitan should be sufficient

Lithrun (changes requested):

Assets/Scripts/Characters/Titan/MindlessTitan.cs
mash8: This is actually necessary for the fix to work. Without it the player titan will still slide around.

Lithrun: See https://github.com/AoTTG-2/AoTTG-2/pull/427/commits/2f5fddc323807e5d37321241ce5e851964f0638a for a solution where PTs will no longer move, and without affecting the previous logic

PR #437, opened 2021-01-02
Time of Day & Timecycle system.

atirutw:
Made to be as easy to use as possible.

Small summary:
The system have 2 parts; the Time of Day which deal with sun & moon and the Timecycle which contain lighting settings for different times of day.

Time of Day:
Controls the sun & the moon.
image
(The script gets the sun's transform from the scene's lighting settings, so you'll have to assign the sun or modify the script)

Timecycles:
Gradient-based workflow. Use gradient to define the color of sun & moon,  ambient lights and fogs(can be created right click > create > timecycle profile).
image
The alpha channel of the gradient is used as multiplier for sun & moon's light intensity and fog density.

### DIFF
--- a/Assets/Scripts/Time of Day.meta
+++ b/Assets/Scripts/Time of Day.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 422df129884d2e24cb068b5f2d987b71
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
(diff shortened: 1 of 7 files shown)


FAStre:
Hi I'll test this out later today, from the screenshots you sent on AOTTG2's discord, it would definitely be an upgrade to the cosmetics of #113 .  Consider joining the Dev Team! The time and lighting system does have more AOTTG type changes I've thought of that would hopefully be added in future, so there is more lighting work to be done if you are very keen to work on lighting, additionally if you dont have much time, there is of course also alot of bug issues to work on aswell.

FAStre:
Had a full look now, looks good! I think most of these changes would come in useful in an upcoming expansion issue however, but including this in 113's branch I think is fine. The build failing seems to be unity errors that there certain script values arnt assigned in the editor, but the build does run, the errors can be ignored(but I will make changes so 113 does pass build by assigning the default values). @Lithrun I think it is fine to approve the pull to 113, where I'll make changes to the scripts to prevent the build failing check.

PR #449, opened 2021-01-06
Removed error messages when entering Chat with HTML Tags

Toorah:
The Problem was, that the fonts weren't marked as dynamic. Not sure why it would break the system as a whole.
This has been resolved.
image

I also modified the error message to give the player an idea why the message is "wrong" as in, incorrect html tags.
image

Lastly, I disabled "rich tags" in the input field, so that the player can see what they are typing.
image


Lithrun (approved):
Seems to be working as intended. Nice 👍

PR #459, opened 2021-01-10
#349 death scoreboard inconsistencies

### LINKED ISSUE #349: [BUG] Death/Scoreboard Inconsistencies
**Describe the bug**
Depending on which way you die, there will either be a message in the kill feed or not, additionally, any "death method" that doesn't appear in the feed, does not update your death counter on the scoreboard.

Normal titans' double-swing attack and Jumpers' leaping bite attack seem to be the only attacks to cause the message to appear in the kill feed (shown as you killing yourself).
Deaths from Annie also show up in the kill feed and update the death counter, but these are different as it says "Female Titan" has killed you instead.

**To Reproduce**
1. Go into a lobby
2. Experience the cold embrace of death
3. Look at kill feed and scoreboard

**Expected behavior**
Regardless of how you die, the death counter on the scoreboard should increase by 1, and all titans should have appropriate kill messages in the kill feed.

**Video**
https://youtu.be/snXpmnlPPDw

**Contributor instructions**

8veStory:
Closes #349. Fixes do the following:

**1. Kill feed info no longer shows in single player.** I noticed the OG game doesn't show kill feed info in single player, which makes sense. I wonder why it occurred in AoTTG2. Example photo of the kill feed showing in single-player to illustrate:

image

**2. The 'GUEST killed GUEST' issue is fixed.** This only occurred for titan attacks that used 'HitEntity' (I.e. the double fist combo). The 'viewID' wasn't set to -1, meaning 'hero.netDie()' uses the player's Photon Player ID instead of the titan's name for the killer's name in the kill feed. I also decided that if the titan is a player titan then viewID will not be forced to -1, so the kill feed shows the player's username instead as the killer.

**3. Deaths in the single player 'Kills/Deaths' feed is now correctly updated for all titan attacks**.

**4. Refactoring:**
- Got rid of duplicate RPCs (OnNapeHitRPC, UpdateHealthLabelRPC)

- Fixed the ambiguous 'HeroHit' event. Added a 'HeroKill' event as part of this refactor.

### DIFF
--- a/Assets/Scripts/Characters/AHSSShotGunCollider.cs
+++ b/Assets/Scripts/Characters/AHSSShotGunCollider.cs
@@ -88,7 +88,7 @@ public class AHSSShotGunCollider : MonoBehaviour
                         var mindlessTitan = item.transform.root.GetComponent<MindlessTitan>();
                         if (PhotonNetwork.isMasterClient)
                         {
-                            mindlessTitan.OnNapeHitRpc(transform.root.gameObject.GetPhotonView().viewID, damage);
+                            mindlessTitan.OnNapeHitRpc(transform.root.gameObject.GetPhotonView().viewID, damage, new PhotonMessageInfo());
                         }
                         else
                         {
@@ -128,7 +128,7 @@ public class AHSSShotGunCollider : MonoBehaviour
                             {
                                 GameObject.Find("MainCamera").GetComponent<IN_GAME_MAIN_CAMERA>().startSnapShot2(item.transform.position, num7, null, 0.02f);
                             }
-                            item.transform.root.GetComponent<FemaleTitan>().OnNapeHitRpc2(base.transform.root.gameObject.GetPhotonView().viewID, num7, new PhotonMessageInfo());
+                            item.transform.root.GetComponent<FemaleTitan>().OnNapeHitRpc(base.transform.root.gameObject.GetPhotonView().viewID, num7, new PhotonMessageInfo());
                         }
                     }
                     else if ((item.transform.root.GetComponent<ColossalTitan>() != null) && !item.transform.root.GetComponent<ColossalTitan>().hasDie)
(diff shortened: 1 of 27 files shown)


Assets/Scripts/Characters/AHSSShotGunCollider.cs
```diff
@@ -88,7 +88,7 @@ private void OnTriggerStay(Collider other)
                         var mindlessTitan = item.transform.root.GetComponent<MindlessTitan>();
                         if (PhotonNetwork.isMasterClient)
                         {
-                            mindlessTitan.OnNapeHitRpc(transform.root.gameObject.GetPhotonView().viewID, damage);
+                            mindlessTitan.OnNapeHitRpc(transform.root.gameObject.GetPhotonView().viewID, damage, new PhotonMessageInfo());
```
Lithrun: I'd recommend to change the TitanBase.OnNapeHitRpc slightly so we don't have to add "new PhotonMessageInfo()" everytime. With the change, it now becomes ` public virtual void OnNapeHitRpc(int viewId, int damage, PhotonMessageInfo info = new PhotonMessageInfo())`, so when no PhotonMessageInfo parameter is passed, it is given the default value.

Assets/Scripts/Characters/EnemyCheckCollider.cs
```diff
@@ -72,14 +72,14 @@ private void OnTriggerStay(Collider other)
...
-                            objArray2[3] = titanName;
-                            objArray2[4] = true;
-                            component.transform.root.GetComponent<Hero>().photonView.RPC("netDie", PhotonTargets.All, objArray2);
+                            object[] netDieParameters = new object[5];
+                            netDieParameters[0] = (Vector3) (((vector5.normalized * b) * 1000f) + (Vector3.up * 50f));
+                            netDieParameters[1] = this.isThisBite;
+                            netDieParameters[2] = myOwnerViewID;
+                            netDieParameters[3] = titanName;
+                            netDieParameters[4] = true;
+                            component.transform.root.GetComponent<Hero>().photonView.RPC("netDie", PhotonTargets.All, netDieParameters);
```
Lithrun: Nice, although when you are changing any RPC which still uses a string, please use `nameof()` instead. See: https://github.com/AoTTG-2/AoTTG-2/wiki/Code-&-Style-Conventions

Lithrun (changes requested):
Looks good. Just two minor comments (and see the sonarcloud)

Lithrun (approved):

PR #466, opened 2021-01-12
#196 add missing audio

### LINKED ISSUE #196: Adding Missing Audio
Find and add all audio not implemented into #98 

- [x] Footsteps

8veStory:
Closes #196. Added all footstep noises as mentioned in the issue.

_Side Note: I personally think that the footstep audio clip is a bit too quiet. Even at max volume the audio clip is underwhelming and doesn't match the sound for all titan sizes (it only really matches the smol ones). It's jarring when you can only hear a large titan walking only if you're quite close to it. If we had a louder footstep noise it would solve this, and since it's louder we would have the leeway to add a feature where the larger the titan is the louder the footstep, so the volume matches the titan size._

_I tried to use Audacity to amplify our current sound to explore this idea but the audio gets garbled and grainy by amplifying it. If we had a clear footstep audio that's louder that'd be cool (preferably one that matches the volume of the other audio clips for consistency). If an audio clip is loud we can always tone it down in the code, but if it's too quite there's not much we can do unless we want to sacrifice audio quality by amplifying._

_That's just my 2 cents though, you might think the footsteps sound fine as is._


Lithrun (changes requested):
Looks nice. Clever that you are using animation events instead of relying on hardcoded animation values.

Only comment I have is that instead of transform.Find, a reference from TitanBody should be used. No need to keep looking for the sound effect GameObject everytime we call this Footstep method, we only need to assign the reference once

Lithrun (approved):

PR #473, opened 2021-01-17
Minor cleanup for IN_GAME_MAIN_CAMERA.cs

Rekperion:
Found some duplicate code in the setMainObject class. Getting rid of this might add a bit of performance boost on lower end machines, cleans up the code a bit, and will reduce the time complexity of the function. I tested on multiple maps to make sure there was no issues generated by removing the code and I have found no bugs generated by this change.

### DIFF
--- a/Assets/Scripts/UI/Camera/IN_GAME_MAIN_CAMERA.cs
+++ b/Assets/Scripts/UI/Camera/IN_GAME_MAIN_CAMERA.cs
@@ -236,14 +236,8 @@ public class IN_GAME_MAIN_CAMERA : MonoBehaviour
     {
         float num;
         this.main_object = obj;
-        if (obj == null)
-        {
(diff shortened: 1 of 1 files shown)


Lithrun (changes requested):
Whereas this change does indeed cleanup cleanup one if statement, it would rather reduce performance instead of increasing it. An if statement isn't considered as an expensive C# operation.

This method `setMainObject` is used 8/ 21 times where its first parameter is null. Instead of assigning the expected, yet duplicate code, it now needs to go twice through a very expensive transform.Find. This is of course, assuming it works.

The `main_object` field is assigned with a null value now. This means that the if statement at line 240 will throw a NullReferenceException. You can test this by entering, and then leaving spec mode. The NullReferenceException is thrown, and you have softlocked yourself and need to rejoin to fix it.

Issue #462, opened 2021-01-11
[BUG] FPS Limiter no longer works - Alpha 2.1


Acros5542:
**Describe the bug**
Seems since v-sync has temporarily become something that is always on, it's taking priority over the FPS Limiter option, so no matter what you put in for the FPS Limiter, your FPS will still be locked to your monitor's refresh rate.

**To Reproduce**
1. Go into any game
2. Open the menu with escape and under graphics settings set a number for the FPS Limiter and hit save
3. Notice the FPS cap doesn't change

**Expected behavior**
Actually, this is sort of behaving how it should - normally in AoTTG if both V-sync and an FPS limit are set, V-sync will take priority. However, right now as V-sync is permanently on this prevents setting an FPS limit at all.

**Additional context**
Right now I see this as extremely low priority, as once V-sync becomes a toggleable option again it will effectively have to be reverted to once again match RC. Unless it's a problem for specific testers with hardware limitations, this one imo isn't worth fixing, just figured I'd make it known.

**Contributor instructions**
1. Try to recreate the bug via the reported reproduction steps. If the bug can be reproduced, remove the investigation label. If the bug cannot be reproduced, please add a comment, attach "On Hold" label and inform the submitter that the bug cannot be reproduced. If there is no reply after 2 weeks, the issue can be closed.
2. Follow the default development team procedures (Create a new branch, apply your fix, create a pull request, submit new builds for testing team in dev squad)
3. Once testing team has confirmed the issue is resolved and the pull request is accepted, the issue can be closed.

Rekperion:
I swear it not stale. Just waiting for review. My PR fixes this issue as well as #352

Lithrun:
Will schedule the review for tomorrow 👍

Rekperion:
Just waiting for another review for PR #474

Issue #352, opened 2020-12-09
[BUG] FPS limit does not persist when changing lobbies


Acros5542:
**Describe the bug**
Regardless of what your setting is for an FPS limit in game, the first server you make when launching the game will always have an FPS limit tied to your refresh rate. Then, for all games you join after this first lobby, the FPS limit is uncapped. You can still cap the FPS limit for a specific lobby by setting the FPS limiter option to something after joining the lobby, but once you leave and join another lobby, the FPS will become uncapped again.

**To Reproduce**
1. Launch the game and join/create a server
2. Check the FPS
3. Leave the server and join/create a new one
4. Check the FPS again
5. Set an FPS cap using the FPS limiter option
6. Leave the lobby and join a new one
7. Check the FPS one last time

**Expected behavior**
The FPS limit option should persist when set, and if left empty, the first game you join when launching the game should have an uncapped FPS limit.

**Video**
https://youtu.be/yNCFzDl6rjs

**Log file**
https://www.dropbox.com/s/bxrgwiguh7km41a/FPS%20limit.txt?dl=0

**Contributor instructions**
1. Try to recreate the bug via the reported reproduction steps. If the bug can be reproduced, remove the investigation label. If the bug cannot be reproduced, please add a comment, attach "On Hold" label and inform the submitter that the bug cannot be reproduced. If there is no reply after 2 weeks, the issue can be closed.
2. Follow the default development team procedures (Create a new branch, apply your fix, create a pull request, submit new builds for testing team in dev squad)
3. Once testing team has confirmed the issue is resolved and the pull request is accepted, the issue can be closed.

Lithrun:
Might be related to either: #351 or #344

Rekperion:
The fix I have up resolves this when it goes through. Won't let me link this issue to my pull for some reason tho.
Edit: It let me link it now :)

Rekperion:
Not stale. Just waiting. Should be all good soon!

Rekperion:
Just waiting for another review for PR #474

PR #474, opened 2021-01-18
#462 & #352 Fixed FPS limiter

Rekperion:
FPS limiter works as intended. Uses VSync when not given an FPS limit as well as fixes several bugs with the FPS limiter that was causing specific input values to make the limit uncapped. FPS limit now also persists into menus once set and is constant for the time the game is running. Unfortunately don't have it so the FPS limit set is saved as I don't know how the prefabs work and I don't want to hold up this issue going through when it can be very beneficial for testers on lower end systems.


Rekperion:
First edits seemed to work when testing through unity but after I made a build the issue was still present.

Rekperion:
After further investigation it appears this fixes it only when not playing full screen. This means the issue could stem from the full screen settings as well as several other issues could only be unique to the full screen view.

Rekperion:
After my personal testing it appears that the FPS just doesn't want to update sometimes and other times it will update all the time without fail. From my testing if the fps cap worked once it will continue to work no matter the change but other times it doesn't update. Still need to investigate where the issue lies but the FPS limiter itself at this point works like a beauty and can be easily modified in the future if needed for new features or changes.

Rekperion:
Much more personal testing and I have found that another system is taking priority over the FPS limiter in the builds as everything works wonderfully in a unity testing environment where as with a build the fps limit doesn't update even if the resolution is changed. Odds are whatever is in control of the VSync is causing this issue.

Rekperion:
The issue was the FengManager having the hard code to always go off the system specs. I have it so the FPS Limiter will default to the system specs if the user does not otherwise have an FPS limit set

Rekperion:
Going through and making tweaks based off tester feedback now. Should be fully done soon.

Rekperion:
SOON TO BE DONE!

Rekperion:
My bad on the bad branch name. I'll have it more streamlined for all work moving foreword.

Lithrun:
Hm, not too sure why the CI keeps failing here. Will try to merge development into this branch, and then see if things still go wrong

Lithrun:
Ah wait, nvm, it's because it's a fork

Assets/Scripts/FengGameManagerMKII.cs
```diff
@@ -2741,9 +2741,6 @@ public void spawnPlayerAtRPC(float posX, float posY, float posZ, PhotonMessageIn
         
         private void Start()
         {
-            QualitySettings.vSyncCount = 1;
```
Lithrun, line 2744: Removing these lines of codes causing the game to run without a FPS limit when you are starting the game. This caused #351

Lithrun (changes requested):
See the review comment. Other than that, try to reduce the amount of comments you are using on your code. Always use /// summary documentation instead of a comment above a method, and also ask yourself if this comment is really necessary. "SetFPSLimit" would mean that this method sets the Frame Limit, so the comment above it is redundant.

Also avoid using built-in type structs, and instead use their C# keyword (i.e. Instead of Int32, use int. Instead of String, use string. Instead of Boolean, use bool ect.)

Assets/Scripts/FengGameManagerMKII.cs
Rekperion: should be good to go now!

Rekperion: I only got rid of the "QualitySettings.vSyncCount = 1;" reference as this will permanently lock the fps with VSync, Which resulted in the limiter not being able to be used at all. Setting the fps upon start in the manager is fine on it's own as this does not lock the FPS and block the FPSLimiter from working.

Rekperion:
I will make sure these changes work on the latest version as soon as I can!

Rekperion:
Needs to be updated with new changes. Will be done in the next couple of days.

Rekperion:
update to PR will be coming thursday/friday

Issue #223, opened 2020-07-03
Melee Weapon Trail


xXXConkerXXx:
**How will this enhancement improve the game?**
This enhancement should fix the old X-WeaponTrail issue.

**Describe the solution you'd like**
Hopefully this should solve our WeaponTrail problems.

**Describe alternatives you've considered**
The alternative is creating the WeaponTrail by scratch.

**Additional context**
https://assetstore.unity.com/packages/tools/particles-effects/melee-weapon-trail-1728

Lithrun:
Issue was already merged as per #701

Issue #33, opened 2020-03-09
Character Customization Phase 1: Setup


Lithrun:
**Is your feature request related to a problem? Please describe.**
Character Customization requires an overhaul. Internally it's structure is odd and doesn't always makes sense. A new character customization screen is requested. This issue focuses on setting up a code infrastructure.

**This requires complicated refactoring of HERO, HERO_SETUP and CustomCharacterConverter. This issue will not be easy! Please consult @Lithrun and define an appropriate code structure with him.**

**Describe the solution you'd like**
A structure which can support the new features.

**Describe alternatives you've considered**
The current structure is too highly coupled. Hard to work it and cannot support new features.

The solution must contain the following:
- [ ] SOLID compliant
- [ ] Support characters having one or multiple skills
- [ ] Remove hard coded set limitations, and allow users to create as many sets as they want. Sets should thereby be able to get imported / exported.
- [ ] Allow to store a EquipmentSet per set (Before spawning you can still change that, but not relevant for this issue)
- [ ] Custom skin URLs must be saved within the character set itself
- [ ] Each skin element must hold a Color property, which is used in phase 2 so that players can recolor a specific part within their skin if they wanted to.
- [ ] Support the 4 human stat attributes (Acceleration, Speed, Equipment and Gas)

WayneZhenLi:
Note to self:  First step is to fix the Hero state machine.  This will allow me to easier include new skills.

Lithrun:
8 Issues pending on this one... Thousands of lines of code to refactor... Now this is where the fun begins (https://www.youtube.com/watch?v=KAHLwAxS7FI)

Lithrun:
For the Github readers:
Progress on Issue 33 is going well so far. Character customization is available again (Being able to change this via a UI #34) and Hero skills are being abstracted. Hero skills will have 3 methods which are called: Use (called when the special attack key is executed. Contains logic on whether or not this skill can be activated), OnUpdate (mainly used to determine whether or not a skill is finished or not), OnFixedUpdate (more complex skills which require physics are using this one)

Also discovered several exploits and cases of sending RPCs per frame, so, after some proper testing and further refactoring, this final AoTTG2 fundamental is ready

PR #495, opened 2021-01-28
#223 weapon trail

mash8:


Lithrun:
This PR does modify several parts of #33. Will have to place this On Hold until #33 has been merged into master

PR #551, opened 2021-02-12
Added new Gravity setting

### LINKED ISSUE #363: New Setting: Gravity
**How will this feature improve the game?**
With the new setting "Gravity", a master client can set the gravity level which applies to all players.

Introduction of a new settings class called "GlobalSettings". This GlobalSettings class contains a property called "Gravity". This property must be of type "float?" (Nullable float). It should act as a multiplayer based on the default AoTTG2 gravity level. Where 1.0 is default, 0.0 would be no gravity and 2.0 would be double the gravity (ect.).

- [ ] Implement gravity (https://docs.unity3d.com/ScriptReference/Physics-gravity.html)
- [ ] Implement GlobalSettings class (Assets.Scripts.Settings)
- [ ] Extend GameSettings to support GlobalSettings
- [ ] Extend SettingsService to support GlobalSettings (Add OnGlobalSettingsChanged(GlobalSettings settings) event)
- [ ] Subscribe to this event in the class where the gravity logic is executed, so whenever the MC updates the gravity, this is changed for everyone.

mash8:
Adding new Global Settings class
Currently has one setting "Gravity"
Changing setting gravity will effect players gravity


Assembly-CSharp.csproj
```diff
@@ -494,66 +496,66 @@
     <Compile Include="Assets\Scripts\Events\Args\HeroHitEvent.cs" />
   </ItemGroup>
   <ItemGroup>
-     <None Include="Assets\Resources\shaders\Unlit - Depth Cutout.shader" />
```
mash8: Honestly don't know what i did to make this happen,

Assets/Scripts/Characters/Human/Hero.cs
```diff
@@ -84,7 +84,8 @@ public class Hero : Human
     private float flareTotalCD = 30f;
     private Transform forearmL;
     private Transform forearmR;
-    private float gravity = 20f;
+    private float Gravity => 20f * gravityModifier;
+    private float gravityModifier = GameSettings.Global.Gravity ?? 1;
```
mash8: Ticket said that the Gravity setting had to be nullable, would it be an idea to just give it a default value of 1 so we don't end up having to do null checks like this every time we want to read it?

play-k: wait u can make a custom map and have gravity like the moon? that's poggers

Lithrun:
#33 has been merged into `origin/development`. @mash8 please resolve the merge conflicts, and then this one should be ready to get merged as well

Lithrun (changes requested):
See https://github.com/AoTTG-2/AoTTG-2/pull/551#issuecomment-798911087

mash8:
> #33 has been merged into `origin/development`. @mash8 please resolve the merge conflicts, and then this one should be ready to get merged as well

Should be good to go now.

PR #534, opened 2021-02-04
#508 Namespace on levelbuilder and block PTs on certain maps

merinobruno:
LevelBuilder now has namespace.
Needed to import Levelbuilder route to a few files:
-ServerSettingsPage.cs
-CreateRoom.cs
-SinglePlayer.cs

Player Titans now are unable to spawn in:
-Akina
-Cave Fight
-House Fight

Issue #508, opened 2021-01-31
[BUG] Player Titans should not be able to spawn on House Fight, Cave Fight and Racing Akina


Acros5542:
**Note**
This is part of a series of post about various error messages that can be given in the Developer Console while playing. Most of these seemingly do not have an impact on gameplay itself, but I am unsure if they are having any significant effect behind the scenes.

**Describe the bug**
Related to #502 as this wouldn't be possible if it weren't for that issue, but if you attempt to spawn as a Player Titan on House Fight, you will be given a IndexOutOfRangeException error - likely because either your model or camera is spawning out of bounds of the map.

**To Reproduce**
1. Go to House Fight
2. Try to spawn as a Player Titan
3. Look at the Developer Console

**Expected behavior**
If it's going to be allowed for you to attempt to spawn as a Player Titan on House Fight, then no error should be occurring and you should be able to spawn in as a Titan on the map.

**Log file**
https://www.dropbox.com/s/615tcsxfex9c4rw/Spawning%20as%20PT%20on%20House%20Fight.txt?dl=0

**Contributor instructions**
1. Try to recreate the bug via the reported reproduction steps. If the bug can be reproduced, remove the investigation label. If the bug cannot be reproduced, please add a comment, attach "On Hold" label and inform the submitter that the bug cannot be reproduced. If there is no reply after 2 weeks, the issue can be closed.
2. Follow the default development team procedures (Create a new branch, apply your fix, create a pull request, submit new builds for testing team in dev squad)
3. Once testing team has confirmed the issue is resolved and the pull request is accepted, the issue can be closed.

Lithrun:
The Player Titan setting should be disabled on this map inside the LevelBuilder. Please also move the LevelBuilder into a namespace.

Apply the same to the Cave Fight map.

Lithrun:
As per title, also include this for Akina.

PR #570, opened 2021-02-21
#508 prevent p ts spawn in certain maps

Lithrun:
Continued from #534


Lithrun:
CI does fail, whereas it does not fail on development. Only original file changes were changes to the LevelBuilder, so I don't know why it fails. @merinobruno perhaps re-apply the same changes in a clean branch?

Lithrun:
New PR was used (#572) . This one can thus be closed

PR #578, opened 2021-02-25
#33 hero refactor toorah v2

Toorah:
- Refactored many variable names
- Refactored RPC use of hard coded strings to proper nameof(...) variation
- split hero class as a partial class (step 1 for long term separation of hero class)
- externalised specific settings or references to serializable container classes
- organised code to some degree
- Cache of GetComponent calls or use of TryGetComponent where possible
- changed instantiation code to use generic overloads
- simplified Hook UI (no more moving disabled UI 10000 units off screen, and colouring a single UI element, instead of using 2)
- Refactored Method names to match Convention()


Lithrun:
Since 33 has been merged, this PR can be closed.

PR #581, opened 2021-02-26
#562 Pause the game correctly while Unpausing

### LINKED ISSUE #562: [BUG] Breaking /pause is my passion
**Describe the bug**
Since the old pause bugs have been fixed, here's a new one. Note, technically this issue doesn't exists on any official Alpha builds yet since none have been made with the /pause fixes, but these fixes are present in some of the individual issue builds which is where I tested this.
This one is pretty simple though - you can no longer re-pause the game. Previously and in AoTTG1, when you the timer for the game to unpause is ticking down (after pausing then unpausing), you could type /pause again to re-pause the game. Since the pause fixes, this now simply resets the unpause timer back to its starting point.

**To Reproduce**
1. Create a multiplayer server
2. Type /pause
3. Type /pause again to being the countdown timer for the game to unpause
4. Before this timer reaches 0, type /pause again
5. Notice it doesn't actually pause the game, but just restarts the timer

**Expected behavior**
Using /pause while the unpause timer is ticking down should pause the game.

**Screenshots**
https://youtu.be/4Ey3M809MLQ

**Contributor instructions**

merinobruno:
Closes #562

### DIFF
--- a/Assets/Scripts/Services/PauseService.cs
+++ b/Assets/Scripts/Services/PauseService.cs
@@ -30,11 +30,19 @@ namespace Assets.Scripts.Services
                 Time.timeScale = 1E-06f;
                 OnPaused?.Invoke(this, EventArgs.Empty);
             }
-            else if (!shouldPause && isPaused)
+            else if (!shouldPause && isPaused && !isUnpausing)
             {
                 isUnpausing = true;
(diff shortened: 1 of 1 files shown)


Lithrun (approved):
Bug Appears to be resolved with above fix. Nice

Issue #568, opened 2021-02-20
[BUG] Racing timers and average speed still update while the game is paused


Acros5542:
**Describe the bug**
More pause bugs! In this one, Racing Akina's (and other racing modes') timers still update when the game is paused with either Escape in singleplayer or /pause in multiplayer. The timers affected include the Race Start countdown, the Time counter indicating how long the race has been going for, and the Average Speed calculation (it will continually average in the speed you were going at the time of the pause).

**To Reproduce**
1. Go to Racing Akina in single or multiplayer
2. Pause the game with either Escape or /pause
3. Look at the counters

**Expected behavior**
These counters should pause when the game is paused.

**Videos**
Singleplayer: https://youtu.be/N18Ozgcy4iI
Multiplayer: https://youtu.be/PlzkdE2OK-8

**Contributor instructions**
1. Try to recreate the bug via the reported reproduction steps. If the bug can be reproduced, remove the investigation label. If the bug cannot be reproduced, please add a comment, attach "On Hold" label and inform the submitter that the bug cannot be reproduced. If there is no reply after 2 weeks, the issue can be closed.
2. Follow the default development team procedures (Create a new branch, apply your fix, create a pull request, submit new builds for testing team in dev squad)
3. Once testing team has confirmed the issue is resolved and the pull request is accepted, the issue can be closed.

PR #586, opened 2021-03-04
Possible fix for the #568 timer issue in racing game mode

xxviiihours:
Fixed issue by disabling Gamemode component whenever pause event is trigger.


Lithrun (changes requested):
Disabling the entire Gamemode doesn't sound like the right solution, and this can cause some issues down the line. Let the Racing Gamemode subscribe to the OnPaused & OnUnPaused events, and then execute some logic to pause / unpause the racing timer.

PR #590, opened 2021-03-10
Fixed objectives not being added to the list properly and fixed assig…

### LINKED ISSUE #342: Racing Objectives Fix
As of the current alpha version, the racing objectives have been completely broken. The problem has to be investigated, and fixed eventually.

LehmannLoic:
…nment of NextObjective

Should fix the issue with racing objectives; The problem was that the RacingGamemode script which is supposed to assign the next objective for each objective was run before the objectives were adding themselves to the list.

The fix works by moving some of code present in OnLevelWasLoaded to new methods AddObjective and AllObjectivesAdded. AddObjective is called in the start method of each RacingObjective.

I could probably cache the FindObjectsOfType<RacingObjective>() used in the comparison as it seems like the objectives are just static objects already placed in the Scenes.

Also added a null check for the audioSource when colliding with Objectives as it is not set in the prefab


Assets/Scripts/Gamemode/Racing/RacingObjective.cs
```diff
@@ -77,8 +77,8 @@ private void Start()
                 Destroy(gameObject.transform.parent.gameObject);
                 return;
             }
-            Gamemode.Objectives.Add(this);
-            State = ObjectiveState.Queue;
+	        State = ObjectiveState.Queue;
+	        Gamemode.AddObjective(this);
```
Rithari: Any particular reason these are swapped?

Rithari (approved):
Looks good to me so far!
Just the question about the two variables being swapped, which is not an issue but I'm curious.
The way the objectives are being handled seems a bit odd, but I don't know enough about the field to know whether or not there is a more efficient way for them to be handled.

Assets/Scripts/Gamemode/Racing/RacingObjective.cs
LehmannLoic: The AddObjective method has a check to call AllObjectivesAdded. AllObjectivesAdded defines the State of the first RacingObjective in the list as the Current objective.
"State = ObjectiveState.Queue;" defaults the State value so if it is not swapped with "Gamemode.AddObjective(this);" all the objectives end up in the "Queue" state.

Rithari: Thanks for clarifying!

LehmannLoic: I do agree that it is a bit odd, the Objectives are not "found" by the RacingGamemode script, they register themselves to RacingGamemode, I didn't change that to not alter the code to much.

Lithrun (changes requested):
Would make more sense indeed if the Gamemode would look for RacingObjectives, rather than these trying to register themselves. So, within OnLevelWasLoaded, you could just do FindObjectsOfType<RacingObjective> and that should be it. _Using this within AddObjective won't scale well_

Additionally, going through the last RacingObjective item does not trigger the "Race Ended", so game does not display what your time was, nor will it ever restart.

LehmannLoic:
> Would make more sense indeed if the Gamemode would look for RacingObjectives, rather than these trying to register themselves. So, within OnLevelWasLoaded, you could just do FindObjectsOfType and that should be it. _Using this within AddObjective won't scale well_
> 
> Additionally, going through the last RacingObjective item does not trigger the "Race Ended", so game does not display what your time was, nor will it ever restart.

Alright I'll work on that, i though the issue was only linked to fixing the objectives themselves.
 I'll make the changes to the retrieving logic of the objectives.

Concerning the Gamemode not ending I found out that some methods in FengGameManagerMKII class are marked as Obsolete (multiplayerRacingFinsih, netRefreshRacingResult refreshRacingResult2 and a few more variables). 
It doesn't seems like that everything marked as obsolete in the FengGameManagerMKII  class has been migrated yet.
Should I work on migrating everything related to RacingObjective in the FengGameManagerMKII class to the gamemode class?

Lithrun:
> 
> 
> > Would make more sense indeed if the Gamemode would look for RacingObjectives, rather than these trying to register themselves. So, within OnLevelWasLoaded, you could just do FindObjectsOfType and that should be it. _Using this within AddObjective won't scale well_
> > Additionally, going through the last RacingObjective item does not trigger the "Race Ended", so game does not display what your time was, nor will it ever restart.
> 
> Alright I'll work on that, i though the issue was only linked to fixing the objectives themselves.
> I'll make the changes to the retrieving logic of the objectives.
> 
> Concerning the Gamemode not ending I found out that some methods in FengGameManagerMKII class are marked as Obsolete (multiplayerRacingFinsih, netRefreshRacingResult refreshRacingResult2 and a few more variables).
> It doesn't seems like that everything marked as obsolete in the FengGameManagerMKII class has been migrated yet.
> Should I work on migrating everything related to RacingObjective in the FengGameManagerMKII class to the gamemode class?

I'd rather recommend to first create the fix so this issue can be resolved. Afterwards, you can create a follow up issue which will refactor any RacingGamemode logic from FengGameManagerMKII into classes where it belongs

Assets/Scripts/Gamemode/Racing/RacingObjective.cs
```diff
@@ -58,7 +58,7 @@ private void SetColor(Color evenColor, Color unevenColor)
             {
                 var materialBlock = new MaterialPropertyBlock();
                 renderers[i].GetPropertyBlock(materialBlock);
-                materialBlock.SetColor("_Color", i % 2 == 0 ? evenColor : unevenColor);
+                materialBlock.SetColor("_BaseColor", i % 2 == 0 ? evenColor : unevenColor);
```
LehmannLoic, line 61: Following the merge of issue #113 the code that updates the colors of the objectives was broken, to fix it I changed the string reference.

LehmannLoic:
I created the new issue to refactor leftover code from FengGameManagerMKII for Racing Gamemode here : #605

Lithrun (approved):
Racing Objectives do work again, however the racing end triggers are not working. That will be resolved as per #605

PR #606, opened 2021-03-25
#462 & #352 Fixed FPS Limiter 2.0

### LINKED ISSUE #352: [BUG] FPS limit does not persist when changing lobbies

### LINKED ISSUE #462: [BUG] FPS Limiter no longer works - Alpha 2.1

Rekperion:
Fixed weird git issues I was having with my old fork and closed it. This is the most recent version and should be all good to go. The main changes are the FPS limiter itself, removal of hard locking the vsyncCount to 1 which was hard locking it into using vsync always, and changes to the QualitySettings.asset that was hard locking the vsync again. Let me know if there is any issues with the update, this is my first time doing a pull straight through VS.

The main thing with this update is defaulting to vsync until a value is input into the fps limiter which it will then switch to. If the bar is cleared at any point it goes back to vsync.


Lithrun (approved):
Unable to reproduce both bugs, so nice

PR #615, opened 2021-04-10
Make Day Night Controller only use one light at a time

### LINKED ISSUE #614: [PROPOSAL] Improvement to sun & moon light
Originally, the Time of Day controller used in 113 was designed to use two directional lights. This was fine for the built-in Render Pipeline(referred to as RP) but a while after we upgraded to Universal RP, I have noticed that the shadow of the two directional lights "snap" at around 6AM/PM in-game. I figured I should update the script to only use one directional light so that the shadows doesn't "snap"

1. Determine if the enhancement is compatible with existing features
2. If the idea is denied, please provide a reason and close the issue. If the idea is accepted, consult your team to do an estimation of effort. Remove the [PROPOSAL] name and label.
3. Move the issue to the Future Milestone. On rare exceptions an issue may be moved to Release if there's sufficient upvotes.
4. Once the issue matches the current milestone, follow development procedures.

atirutw:


Assets/Scripts/DayNightCycle/DayAndNightControl.cs
```diff
@@ -181,22 +181,26 @@ void UpdateLightingSettings()
...
             {
-                DirectionalLight.color = timecycle.sunlightColor.Evaluate(CurrentTimeScale);
-                DirectionalLight.intensity = timecycle.sunlightColor.Evaluate(CurrentTimeScale).a * timecycle.maxSunlightIntensity;
+                Sun.color = timecycle.sunlightColor.Evaluate(CurrentTimeScale);
+                Sun.intensity = timecycle.sunlightColor.Evaluate(CurrentTimeScale).a * timecycle.maxSunlightIntensity;
             }
-            if (timecycle.overrideMoonlight)
+            if (timecycle.overrideMoonlight && isNightTime)
             {
                 Light moonLight = Moon.GetComponent<Light>();
```
Lithrun: So now that Moon is Light already, using GetComponent doesn't make sense here

Lithrun (changes requested):
Small improvement

Lithrun (approved):

PR #708, opened 2021-09-18
#668 vertical hook pull

### LINKED ISSUE #668: [BUG] Vertical hook pull is too weak
**Describe the bug**
When your character hooks any object above or below them, they're pulled upwards or downwards towards that object.  The farther away the object, the harder your character is pulled.  If you land two hooks, the pull is increased substantially.

In AoTTG2, vertical hook pull is significantly weaker than it is in RC.  This is due both to the characters having incorrect ACL values, making them too heavy, and seemingly some further problem in the code--they're pulled less than even characters of their weight should be.

**To Reproduce**
1. Copy this map script and load it into RCMod/Anarchy/Your AoTTG 1 mod of choice in racing mode: https://pastebin.com/eaFnxWwJ
2. Spawn as Eren, who has 100 ACL.
3. Look up at a block above you, so that the reticle reports a distance of 101.  Tap (do not hold) a single hook key.
4. You should be pulled 79-80 units in the air, completing the "race" a single time at the very apex of your jump.  Any character with more ACL than Eren will travel higher, and be registered as completing the "race" twice.
5. Copy this map script and load it into AoTTG 2 in racing mode:  https://pastebin.com/A5MFkNiD
6. Spawn as any 100 ACL character (all of them, at the time of reporting), and repeat step 3.  You will not come even close to finishing.

**Expected behavior**
To put it simply, AoTTG2 should perform *identically* to AoTTG 1 mods.  It does not.  This is a complex issue however, and making them behave the same in only this specific example would be insufficient.

**Additional context**
Currently, it is impossible to empirically collect enough information about the details of this bug to fix it.  #624 must be completed first, before additional testing is possible to identify the exact issue.

A note to save this information where it is visible: According to Order (the developer of Anarchy mod), ACL determines character mass for hook pull by this formula: `mass = 0.5 - (Acl - 100) * 0.001;`
AoTTG2 should match this, though I expect the bug is more complex than a simple formula being wrong.

TiltSword:
Closes #668 

**Notes for the reviewer:**

I confirmed together with antigasp that the entire issue was incorrect character mass values. 
For now, I set the mass of the character to be 0.45, which is equivalent to Levi's mass with his 150 ACL stat. *This is only a temporary fix so testers can work with the ODM better.* Once the character-specific stats are implemented, the mass calculation will need to be adjusted. 

**Contributor guidelines:**

### DIFF
--- a/Assets/Scripts/Characters/Humans/Hero.cs
+++ b/Assets/Scripts/Characters/Humans/Hero.cs
@@ -2097,7 +2097,14 @@ namespace Assets.Scripts.Characters.Humans
                 photonView.RPC(nameof(InitializeRpc), PhotonTargets.OthersBuffered, config);
             }
 
+            //float acl = to be implemented
+            Rigidbody.mass = 0.45f; /*0.5f - (acl - 100f) * 0.001f;         <- according to antigasp and Order, this is the formula for character mass
+            I was asked by antigasp to use 0.45 (corresponding to ACL 150) as a placeholder because most testers are used to playing as Levi and it'd be
+            easier for them to spot if something is wrong. Obviously this is going to have to be reworked once character-speficic stats are implemented,
+            but for now it would probably make life easier for the testers.*/
+
             EntityService.Register(this);
+            Debug.Log("Mass of character: " + Rigidbody.mass);
         }
 
         [PunRPC]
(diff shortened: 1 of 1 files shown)


TiltSword:
About the failed quality gate: it's a piece of code that I commented out which is only going to be usable _after character-specific presets have been implemented_, I wrote it down so it's more easily accessible later on when the aforementioned part is added, even if I'm not going to be the one working on this issue later on.

Lithrun (approved):
LithrunsSealOfApproval

PR #694, opened 2021-08-28
#604 settings overhaul

Lithrun:
Closes #604

**Notes for the reviewer:**

Reworks the entire settings system that was used for AoTTG2 & uses ScriptableObjects

**Contributor guidelines:**
1. A build has been uploaded to the discord #builds channel & has been tested by the testing team
2. Assure the code is inline with the [AoTTG2 coding conventions](https://github.com/AoTTG-2/AoTTG-2/wiki/Code-&-Style-Conventions)
3. Fix any issues that the CI reports (Sonarcloud & unit tests)

After all above conditions are met, @Lithrun will review your Pull Request. Resolve any comments of the reviewer or discuss them if you disagree.

Issue #604, opened 2021-03-24
Level & Settings: ScriptableObject Deluxe Edition


Lithrun:
**How will this enhancement improve the game?**
After the fancyness of #489 and #599, there now comes a sequel to make the triology complete: Convert the GameSettings & LevelBuilder classes into a ScriptableObject. This will allow us to manage these settings more dynamic and easier via the unity editor, without having to modify any code. Useful when you want to test a specific set of settings as well in the Unity Editor.

**Describe the solution you'd like**
ScriptableObjects are fancy (as long as they are used for what they are intended for...), let's assure the Levels and Settings will work as ScriptableObjects too.

Lithrun:
Assigned to self, but will first pickup #645 as it's more urgent.

PR #739, opened 2021-11-08
Fixed Duplication Exploit

### LINKED ISSUE #276: [BUG]#276 Duplication exploit
**Describe the bug**
When a player dies, If you click Item1 to spawn, then after spawning use the /specmode command and then spawn again, a duplicate of your character appears.

**To Reproduce**
1. Open any gamemode with titans
2. Die by titan
3. Click Item1 key(1 by default) and spawn
4. Use the /specmode command in chat and spawn using item1 or 1 key again
5. You now have a duplicate player

**Expected behavior**
No duplication should be occurring at all

PeteAniDev:
Fixed duplication exploit due to incorrect spectator mode state.

Closes #276

**Contributor guidelines:**


PeteAniDev:
Since the issue is resolved, I will not continue working on migrating Fenglee Codes to the new system as this "Code Smell" part is related to Gamemode settings. The migration of this should be completed once needChooseSide instance of Fenglee is migrated.

Lithrun:
> 
> 
> Since the issue is resolved, I will not continue working on migrating Fenglee Codes to the new system as this "Code Smell" part is related to Gamemode settings. The migration of this should be completed once needChooseSide instance of Fenglee is migrated.

That's fine, as Gamemode Settings are significantly changed as per #604 (You can take a peek at #694)

Sonarcloud comments are there as a warning, not as an error. So you can ignore those which aren't relevant to your issue. If there is one which shouldn't be ignored, then you will hear it from me

Assets/Scripts/FengGameManagerMKII.cs
```diff
@@ -1543,7 +1507,8 @@ private void loadconfig()
             objArray[0xf2] = PlayerPrefs.GetString("hjump", "Q");
             objArray[0xf3] = PlayerPrefs.GetString("hmount", "LeftControl");
             objArray[0xf4] = PlayerPrefs.GetInt("chatfeed", 0);
-            objArray[0xf5] = 0;
+            objArray[0xf5] = null; // Is no longer in the code
```
Lithrun: If this is no longer required, then delete it

Assets/Scripts/UI/Camera/SpectatorMode.cs
```diff
@@ -0,0 +1,83 @@
...
+using UnityEngine;
+using static Assets.Scripts.FengGameManagerMKII;
+
+namespace Assets.Scripts.Utility
+{
+
+    public static class SpectatorMode
+    {
+
+        private static bool _disable = true;
```
Lithrun: Rather switch this "disabled" to "enabled", as it's more common to check if something is enabled or not enabled vs disabled or not disabled.

Assets/Scripts/UI/Camera/SpectatorMovement.cs
```diff
@@ -1,11 +1,13 @@
...
+using Assets.Scripts.Utility;
 using UnityEngine;
 
+// Call with GameObject.Find("MainCamera").GetComponent<SpectatorMovement>()
 public class SpectatorMovement : MonoBehaviour
 {
-    public bool disable;
-    private float speed = 100f;
+    public bool disable = true;
+    private readonly float baseSpeed = 100f;
```
Lithrun: A constant (const) would be better here, as this value will not change throughout the usage of the program. With readonly it is technically allowed to modify the value via the constructor, whereas via const you make this more explicitly clear

Assets/Scripts/UI/Camera/SpectatorMovement.cs
```diff
@@ -16,11 +18,12 @@ private void Awake()
 
     private void Update()
     {
-        if (!this.disable)
+        SpectatorMode.Initialize();
```
Lithrun: Continuously calling an "Initialize" method doesn't seem right to do every frame. This will keep setting the "_disable" to true.

Assets/Scripts/UI/Camera/SpectatorMovement.cs
```diff
@@ -16,11 +18,12 @@ private void Awake()
...
+        if (!disable)
```
Lithrun: You could invert this if statement to make the code a bit more readable. Thus the logic wouldn't be nested within the if statement

Compare:
```
if (disable) return;

// All of the movement logic
```

vs

```
if (!disable)
{
     // All of the movement logic
}
```

Lithrun (changes requested):
See comments. Additionally, what is the exact correlation between the "disabled" state of the SpectatorMode and the actual usage of the EnterSpecMode? Perhaps a cleaner approach would be to call EnterSpecMode whenever the Specmode is enabled / disabled, as then spectator mode specific logic can be within the spectatormode class and the only thing that other classes have to do is to then either enable / disable it

Assets/Scripts/UI/Camera/SpectatorMovement.cs
PeteAniDev: Oh, I forgot that thing. It should be in Awake()

PeteAniDev:
The EnterSpecMode() can only be used once everything is initialized. For example, when you press 1 to enter the game you switch out of spectator mode. But your body is not initialized yet so the EnterSpecMode() should run after you spawn your hero.
My Unity crashed when trying to put EnterSpecMode() in Toggle(), Disable(), and Enable().

I'll change its name from EnterSpecMode() to UpdateSpecMode() for clarity.

Lithrun (approved):

PR #743, opened 2021-11-10
#350 Oversized Titan Arm Hitboxes

### LINKED ISSUE #348: [BUG] Oversized Arm Hitboxes
**Describe the bug**
The arm hitboxes on titans seem to either be oversized or reach too high, causing you to slice their arm at times when your blades realistically should never had touched them.

**To Reproduce**
1. Go on any map
2. Swing at nape or higher level of the titans
3. Stick 'em with the pointy end of your blades
4. Notice the times when the arm is sliced as well when it probably shouldn't have been

**Expected behavior**
Arm hitbox should more closely match where the model of the titan's arm is at.

**Video**
If you put it on 0.25x speed at around 11 seconds you can see it pretty clearly.
https://youtu.be/P-icg8C5iMc

**Additional context**
This is likely a lower priority bug since it doesn't affected gameplay that much.

### LINKED ISSUE #350: [BUG] Hit Detection
**Describe the bug**
AoTTG2 seems to be confirming kills on the first FixedUpdate of overlap between your blade hitboxes and the nape, whereas in AoTTG1 it requires two FixedUpdates of overlap (the function OnTriggerStay requires this). By removing this two frame requirement, it drastically changes the scene of high damage kills and effectively removes the theoretical cap on possible damage done with blades. This is not a good thing, effectively turning the entire scene of competitive high damage into a game of RNG rather than skill as most of the fine-tuning and skill required at the highest levels of high damage is based around trying to go as fast as possible while still positioning/angling yourself so that your blades overlap for two FixedUpdates. I'm not an expert on this, and there are other players in the community who can explain how this works and why it's so important much better than me (Dr. Eisenlocke for example), but the kill in the clip below proves AoTTG2 is not requiring this two FixedUpdate limit.

**To Reproduce**
Reproducing proof of this bug requires hitting damage with blades a bit over 6.4k, or at least having a good enough feel for how hit detection should work to tell if a kill lower than that shouldn't have registered when it did (fast moving or mid-animation titans hit for around high 3k to 4k damage can also sometimes show this)

**Expected behavior**
As mentioned above, AoTTG2 should use two FixedUpdates (two frames when on 50 fps) to determine if a kill is confirmed or missed, not one. So for example, in the attached clip, my kill should have missed.

**Video**
https://youtu.be/_tkLzdKMB0s

**Additional context**
I know this might seem like a small difference between the two games, but for the competitive community it is an especially major change.

**Contributor instructions**

Zynku:
Changed the capsule colliders around the arm hitbox to be more consistent with what's expected. A build was already created and tested, I just forgot to commit to the branch.

Closes #348

**Notes for the reviewer:**

Committing changes that were made in a build already. Should be good to go

**Contributor guidelines:**

### DIFF
--- a/Assets/Resources/MindlessTitan.prefab
+++ b/Assets/Resources/MindlessTitan.prefab
@@ -305,9 +305,9 @@ CapsuleCollider:
   m_IsTrigger: 0
   m_Enabled: 0
   m_Radius: 0.25
-  m_Height: 1.4
+  m_Height: 0.6969134
   m_Direction: 0
-  m_Center: {x: -0.25, y: 0, z: 0}
+  m_Center: {x: -0.48155117, y: 0.0000005855481, z: 0.000000015336342}
 --- !u!1 &1114370160899894
 GameObject:
   m_ObjectHideFlags: 0
@@ -3340,9 +3340,9 @@ CapsuleCollider:
   m_IsTrigger: 0
   m_Enabled: 0
   m_Radius: 0.25
-  m_Height: 1.4
+  m_Height: 0.6969134
   m_Direction: 0
-  m_Center: {x: -0.25, y: 0, z: 0}
+  m_Center: {x: -0.48155117, y: 0.0000005855481, z: 0.000000015336342}
 --- !u!1 &1928165841424312
 GameObject:
   m_ObjectHideFlags: 0


Lithrun:
This is actually the fix for #348, not for #350

Lithrun (approved):
In comparison to AoTTG's arm hitboxes and the way we are using them, these seem to fit much better. Which also automatically resolves the issue of cutting both nape and arms at the same time. Nice one

PR #745, opened 2021-11-11
New Art Prefabs/Tweaks + Particles

TonyTwohands:
Hey Lith, not sure how to go about these, but I added a bunch of particles and prefabs I thought would be cool. There's not really much to test as most of these are purely visual and I've tested them myself countless times to make sure they look right.
Basically just a MK II for #730


Lithrun (approved):
LithrunsSealOfApproval
A lot of new fancy stuff. Well done

PR #747, opened 2021-11-15
#179 unresponsive reel in scroll wheel

### LINKED ISSUE #179: [BUG] Reel-in feels unresponsive, especially with default controls

PhantomRyser:
Ensured that reeling inputs are checked in both the Update and FixedUpdate methods. Will most likely change how inputs are read once Hero.cs is refactored properly but this ensures that there isn't any missed inputs and labelled some variables that were ambiguous.

Closes #179

**Notes for the reviewer:**

*(Add information which you think is important for the reviewer to be aware of (i.e. functionality X is not included as this was too complex, see issue 123))*

**Contributor guidelines:**


Assets/Scripts/Characters/Humans/Hero.cs
```diff
@@ -1281,6 +1286,37 @@ public void Update()
...
+                reelForce = -1f;
+            }
+            else if (InputManager.Key(InputHuman.ReelOut))
+            {
+                reelForce = 1f;
+            }
+            else if (InputManager.KeyUp(InputHuman.ReelIn) || InputManager.KeyUp(InputHuman.ReelOut))
+            {
+                reelForce = 0f;
+            }
```
PeteAniDev: I believe you should use `&&` here instead of `||`.
And remove `else` because if one of the two cases up there is true, those lines will never be reached.

PhantomRyser: I don't think that anyone would release the reel in and reel out key binds at the same time which would be what is required to make that if statement true (unless KeyUp always returns true if not pressing the key). I'm not sure but I think it is only true once on the frame that the key is released (or not pressed anymore). 

I realize now that I can get rid of that entire elseif statement because the reelForce will "zero" out at the end of the statement chain if scroll inputs aren't being used.

Lithrun (approved):

PR #756, opened 2021-11-23
#752 Invisible Blade Bug

### LINKED ISSUE #752: Interrupting the reloading animation can result in an invisible blade. [BUG]
**Describe the bug**
The title described its all.

**To Reproduce**
1. Spawn as a hero.
2. Reload.
3. Hook to interrupt the reload.
4. Kill titans with the invisible blade.

**Expected behavior**
Interrupting reload animation should make the hero fail to reload and have to reload again.
The HUD should also display that you haven't reloaded yet.

Zynku:
Closes #752

**Notes for the reviewer:**

Should work as intended. Interrupting a reload animation with blades will now give the player the new blade anyway, and act like the animation wasn't interrupted. If you think the player should not be given the new blade, and should be instead forced to reload again, please state so.

**Contributor guidelines:**


Assets/Scripts/Characters/Humans/Hero.cs
```diff
@@ -449,6 +450,12 @@ public void Update()
                 bool isLeftHookPressed;
                 BufferUpdate();
                 UpdateExt();
+                if (state != HumanState.ChangeBlade && weaponDisabledOnReloading)
+                {
+                    //If the reload animation is cancelled before the weapon has a chance to be reenabled, call this function to do that
+                    Equipment.Weapon.EnableWeapons();
+                }
```
PeteAniDev, line 463: The weapon will be enabled repeatedly since Equipment.Weapon.EnableWeapons() does not change weaponDisabledOnReloading to false. Indeed, I don't see that weaponDisabledOnReloading is ever false. See the next review, it might be a possible solution.

Assets/Scripts/Characters/Humans/Equipment/Weapon/Blades.cs
```diff
@@ -26,40 +27,54 @@ public override void PlayReloadAnimation()
...
+        {
+            if (TotalBlades > 0)
+            {
+                WeaponLeft.SetActive(true);
+                WeaponRight.SetActive(true);
+                Hero.currentBladeSta = Hero.totalBladeSta;
+                bladesThrown = false;
+                Hero.weaponDisabledOnReloading = true;
+            }
+        }
```
PeteAniDev: Since this enables the weapons, shouldn't this set Hero.weaponDisabledOnReloading to false instead of true?

PeteAniDev (changes requested):
Canceling the reload helps you instantly reload is a cool mechanic, but you should wait for Lithrun to approve this idea.

Assets/Scripts/Characters/Humans/Equipment/Weapon/Blades.cs
Zynku: Yeah, I suppose I have the name backwards, should probably be called weaponEnabledAfterReloading, or invert the bool's usage

PeteAniDev: Well even if you change the name, the boolean is still always true, so it might be unnecessary.

Assets/Scripts/Characters/Humans/Hero.cs
Zynku: This function is called only when its any state but reloading. I.e. if we switch to a hook animation, and the weapons are still disabled, then it will be called, which does the "reloading" instantly. It's set to false in the function that throws the blades so that its true in the frames between the blades being thrown (since they disappear then) and the blades being brought of the holsters (they are reenabled here)

Zynku: Wait...

Lithrun (approved):

PR #758, opened 2021-11-26
#757 inconsistent and camera angle dependent fog

### LINKED ISSUE #757: inconsistent and camera angle dependent fog [BUG]
**Describe the bug**
On the medium settings the fog is inconsistent between the ground and other objects. You can see in the left picture that the trees and titan are affected differently compared to the ground plane.

When I turn around the titan (and face an other way) it sometimes seems to render correctly.
picture is the same time of day (I choose one that highlights the issue).

When I turn fog off in the preview or in the render settings the issue disappears completely.

~~A quick fix would be to turn off the fog completely since the fogcolor also does not seems to be synched to the skybox color with the time of day.~~ (see fix below)

**To Reproduce**
1. start a map
2. open game settings and drag the slider to the left (about 25% from left to right), 
3. notice the fog increasing
4. move around the level a bit and notice when you turn the camera 360deg the ground fog flickers
5. verify that the same problem exists in the editor (while the game is running if you move around there you can also see the problem)

**Expected behavior**
the fog stays consistent independent of camera angle, preferably the option on the right

**Screenshots**
image
image

**Log file**
Include a log file

**Additional context**
Did some experiments and seems that this issue is caused by the ground plane that is a box of which the part you see are 2 **HUGE** polygons the fog does not know how to deal with this properly.

Fix by increasing polygon number on rendered object: 
- [x] remove rendering component on the collision box
- [x] add default plane
- [x] scale plane to 175,1,175. move plane around (to fit over the original collision box) and make sure plane y=0
- [x] remove collision on plane

subject546:
Closes #757
**Notes for the reviewer:**
please check if fog is fixed.
please check if nowhere the rendering is not overlapping the collision.

probably a good idea if the @AoTTG-2/mapping-team took a look at this PR.

Notes:
- For future maps it should be noted that the ground can not be a box with 2 **HUGE** polygons as top... see comment in issue #757 
- This Issue might still exist on custom maps! One solution might be to use a subdivided cube instead of a simple one if the scale is bigger than a certain value after which this bug exists.

**Contributor guidelines:**


dark3art:
all new maps use a unity terrain object and not a cube with 2 Huge polygons so don't worry

Lithrun (approved):
Nice one. Did exclude the Forest change you did tho, as the latest forest version no longer has a cuboid, instead it uses a terrain now

PR #761, opened 2022-02-02
#712 Locked camera flip fix

### LINKED ISSUE #712: [BUG]Locking onto a titan and shaking the mouse violently will cause the screen to break
**Describe the bug**
Already an issue in AoTTG1 as well. Locking onto a titan and shaking your mouse violently will cause the screen to get flipped at different angles and also completely upside down as well.

**To Reproduce**
1)Lock onto a titan (default key f)
2)Start shaking your mouse violently and in different directions. It might take a bit of time to do this.

**Expected behavior**
The screen should remain at its original angle and not get funny

**Screenshots**
image

**Log file**
N/A

**Additional context**
This is a carry over bug from AoTTG1.

SaudBako:
Fixes #712 

**Notes for the reviewer:**
I simply stopped moving the camera if the player is currently locked onto a titan.

**Additional notes**
- In addition to fixing the issue, this makes locking a better experience, because the camera is no longer shaking on mouse input.
- I identified the actual culprit, line 580 in 2f018e7d614d7ae00f25a0d63d1396feb2aef4a1, but didn't modify it, as I'm not sure what it does.


Lithrun (approved):
After a lot of violent mouse shaking, it did nothing, thus we can consider this issue complete

PR #763, opened 2022-02-09
Disabled horses in the GamemodeSettings constructor

### LINKED ISSUE #512: [BUG] Getting the horse to follow you across maps/gamemodes
**Describe the bug**
If you go to Outside the Walls, then open the pause menu with escape and switch maps, the horse will follow you to other gamemodes once you've restarted the round.

**To Reproduce**
1. Go to Outside the Walls
2. Spawn in as a Human
3. Press escape
4. Click on Game Settings
5. Change the map to something else (not Trost)
6. Click change
7. Exit out of the menu and type /restart (or T if on singleplayer)

**Expected behavior**
The horse should only appear on gamemodes where it is allowed.

**Screenshots**
Screenshot (526)

**Log file**
https://www.dropbox.com/s/s6px6hgzou3786p/Horse%20across%20maps.txt?dl=0

**Contributor instructions**

HeyImBrian:
Closes #512

I apologize that it took me so long to create a pull request. I got pog reacts to my uploaded build, but I didn't interpret that as confirmation that it works. Future builds from others have only been reacted to, so I assume that's the way it is. 

I disabled horses in the GamemodeSettings constructor because once the horses were set to enabled, they remained enabled no matter which map the player went to.

### DIFF
--- a/Assets/Scripts/Settings/Gamemodes/GamemodeSettings.cs
+++ b/Assets/Scripts/Settings/Gamemodes/GamemodeSettings.cs
@@ -66,6 +66,7 @@ namespace Assets.Scripts.Settings.Gamemodes
                 Mindless = new MindlessTitanSettings()
             };
             Horse = new HorseSettings();
+            Horse.Enabled = false;
             Respawn = new RespawnSettings();
             Time = new TimeSettings();
             TeamMode = TeamMode.Disabled;


Lithrun (approved):
Hm, sounds like we need some proper way on how to use reactions on builds then. As a "pog" doesn't sound like any confirmation either

PR #772, opened 2022-03-24
#765 crosshair adjustment

### LINKED ISSUE #765: [BUG] crosshair adjustment
**Describe the bug**
There's an issue with the crosshair. Left hooks land much lower than expected, whereas right hooks land much higher than expected. Which as you can imagine, makes it quite annoying to test ODMG logic. Issue did not occur before https://github.com/AoTTG-2/AoTTG-2/commit/083f550fd58b0cf09d7c253301ff472776d05f5b, and was caused somewhere during the Pull Requests

**To Reproduce**
1. Aim at something with your hooks
2. Left hook will hit below the crosshair whereas the right hook will hit above the crosshair

**Expected behavior**
It's expected that the hook will land exactly where the crosshair aimed it at

**Screenshots**

**Log file**
Include a log file

**Additional context**

mwnDK1402:
Closes #765

**Notes for the reviewer:**
I have made two changes:
- The left hook should now be at the same exact height as the right hook.
- Both hooks should land exactly where the cursor is pointing (in TPS, of course). Previously, hooks would land ~1.4 units above their intended target.

**Contributor guidelines:**
- [x] A build has been uploaded to the discord #builds channel
- [x] The build has been tested by the testing team
- [x] Assure the code is inline with the [AoTTG2 coding conventions](https://github.com/AoTTG-2/AoTTG-2/wiki/Code-&-Style-Conventions)
- [x] Fix any issues that the CI reports (Sonarcloud & unit tests)

### DIFF
--- a/Assets/Resources/Character2/HumanBase2.prefab
+++ b/Assets/Resources/Character2/HumanBase2.prefab
@@ -17148,7 +17148,7 @@ Transform:
   m_PrefabAsset: {fileID: 0}
   m_GameObject: {fileID: 4759155575018946212}
   m_LocalRotation: {x: 0.06328958, y: 0.06328935, z: 0.70426875, w: 0.70426875}
-  m_LocalPosition: {x: 0.001046091, y: -0.001890445, z: 0.001161152}
+  m_LocalPosition: {x: 0.001020005, y: -0.001890445, z: 0.001175438}
   m_LocalScale: {x: 0.009405191, y: 0.009405194, z: 0.009405193}
   m_Children: []
   m_Father: {fileID: 4755987445250634402}
--- a/Assets/Scripts/Characters/Humans/Bullet.cs
+++ b/Assets/Scripts/Characters/Humans/Bullet.cs
@@ -1,3 +1,4 @@
+using System;
 using Assets.Scripts.Characters.Titan;
 using System.Collections;
 using UnityEngine;
@@ -63,7 +64,7 @@ namespace Assets.Scripts.Characters.Humans
         {
             this.phase = 2;
             this.killTime = 0f;
-            object[] parameters = { 2 };
+            object[] parameters = {2};
             base.photonView.RPC(nameof(setPhase), PhotonTargets.Others, parameters);
         }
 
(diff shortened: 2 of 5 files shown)


Assets/Resources/Character2/HumanBase2.prefab
```diff
@@ -17148,7 +17148,7 @@ Transform:
   m_PrefabAsset: {fileID: 0}
   m_GameObject: {fileID: 4759155575018946212}
   m_LocalRotation: {x: 0.06328958, y: 0.06328935, z: 0.70426875, w: 0.70426875}
-  m_LocalPosition: {x: 0.001046091, y: -0.001890445, z: 0.001161152}
+  m_LocalPosition: {x: 0.001020005, y: -0.001890445, z: 0.001175438}
```
mwnDK1402, line 17151: I attempted to align `hookRefL1` and `hookRefR1`.
It seemed to make a difference. I did not do this for `hookRefL2` and `hookRefR2`, since they're not under the same parent, and I couldn't really compare them the same way.

Assets/Scripts/Characters/Humans/Bullet.cs
```diff
@@ -214,75 +215,46 @@ private void killObject()
             UnityEngine.Object.Destroy(this.rope);
             UnityEngine.Object.Destroy(base.gameObject);
         }
-
-        public void launch(Vector3 v, Vector3 v2, string launcher_ref, bool isLeft, GameObject hero, bool leviMode = false)
+        
+        public void Launch(Source source, GameObject hookRef, Vector3 v, Vector3 v2, Hero hero, bool leviMode = false)
```
mwnDK1402: I'm using a `Source` enum to carry the data about which transform the hook originated from, for example `hookRefL1`. This has the benefit of making it easier to create one authoritative `LaunchRope` method in `Hero`, and it allows us to send that enum rather than a `string` over the network in the `myMasterIs` RPC.

Assets/Scripts/Characters/Humans/Bullet.cs
```diff
@@ -504,6 +479,23 @@ private void Update()
...
+            source switch
+            {
+                Source.BeltLeft => true,
+                Source.GunLeft => true,
+                Source.BeltRight => false,
+                Source.GunRight => false,
+                _ => throw new ArgumentOutOfRangeException(nameof(source), source, null)
+            };
+
+        public enum Source
```
mwnDK1402: Name suggestions are welcome.

Assets/Scripts/Characters/Humans/Hero.cs
```diff
@@ -240,6 +240,9 @@ private HumanState state
         public Rigidbody Rigidbody { get; protected set; }
         public SmoothSyncMovement SmoothSync { get; protected set; }
 
+        private Bullet.Source LeftSource => useGun ? Bullet.Source.GunLeft : Bullet.Source.BeltLeft;
```
mwnDK1402: Only really used in the two `LaunchRope` methods, so I could get rid of these.

Assets/Scripts/Characters/Humans/Hero.cs
```diff
@@ -3207,56 +3203,53 @@ public void Launch(Vector3 des, bool left = true, bool leviMode = false)
...
+                LaunchHook(hook);
                 launchPointRight = Vector3.zero;
             }
+
+            void LaunchHook(Bullet hook)
+            {
+                var num = !single ? distance <= 50f ? distance * 0.05f : distance * 0.3f : 0f;
+                var hookPos = hook.transform.position = hookRef.transform.position;
+                var vector = Vector3.Normalize(point - transform.right * num - hookPos) * 3f;
+                Debug.DrawLine(hookPos + Vector3.up * 0.01f, point + Vector3.up * 0.01f, Color.green, 10f);
```
mwnDK1402: Didn't notice these, I'll get rid of them ASAP.

Assets/Scripts/Characters/Humans/Hero.cs.meta
```diff
@@ -1,11 +1,10 @@
 fileFormatVersion: 2
 guid: c2ade5059944e77468ce66f0f1fb36c9
-timeCreated: 1582403864
-licenseType: Free
 MonoImporter:
+  externalObjects: {}
   serializedVersion: 2
   defaultReferences: []
-  executionOrder: 0
+  executionOrder: 5
```
mwnDK1402, line 7: This part is important, this fixes the hook target offset by ensuring that the camera position is correct, since with default timing, the returned camera position is ~1.4 units too high.

mwnDK1402 (commented):
I need to remove some `Debug.DrawLine` calls.

mwnDK1402:
Waiting for response from the testing squad.

mwnDK1402 (commented):
I don't see any more issues.

Assets/Scripts/Characters/Humans/Hero.cs
```diff
@@ -3207,56 +3203,51 @@ public void Launch(Vector3 des, bool left = true, bool leviMode = false)
...
-                    hookRight.launch((vector * 3f), Rigidbody.velocity, str, false, gameObject, false);
-                }
+                hookRight = hook;
+                LaunchHook(hook);
                 launchPointRight = Vector3.zero;
             }
+
+            void LaunchHook(Bullet hook)
+            {
+                var num = !single ? distance <= 50f ? distance * 0.05f : distance * 0.3f : 0f;
```
mwnDK1402: Code smell: "Extract this nested ternary operation into an independent statement."

mwnDK1402:
@AoTTG-2/testing-team seemed to find no issues.

Assets/Resources/Character2/HumanBase2.prefab
Lithrun: Seems fine

Assets/Scripts/Characters/Humans/Bullet.cs
Lithrun: Perhaps name it something like HookSource would be better, as Source seems too generic

Assets/Scripts/Characters/Humans/Hero.cs
Lithrun: If it's only used within LaunchRope, then convert it into a variable.

Lithrun (changes requested):

Lithrun (approved):

